PR #252, opened 2024-12-03
feat: Improve 404 page navigation

### LINKED ISSUE #248: Improve 404 page display
It would be helpful to display navigation on the 404 page. Design and implement an improvement.

CleanShot 2024-12-01 at 22 14 15@2x

ansen01-svg:
## What changed?
Added logo, button and a link to the 404 page that redirects users to the homepage and last visited page.

## Why?
The original 404 page lacked an easy way to return to the home page or the previous page, requiring manual URL entry or browser navigation.

## How was this change made?
- Used the Empty component and added a button and a link.
- Integrated router.history for home page redirection.

## How was this tested?
- Manually tested the functionality of the 2 links present on the page.

## Anything else?
Adjustments may be needed based on UI feedback or additional device testing.


kydecker:
Hi @ansen01-svg! Thanks for proposing a change here! I realize that ticket #248 didn't include many details, so thanks for diving in here.

Before merging this, I'd want to add a few other things:

1. Can you add the `Logo` to the top of this page, above the `Empty` component? There's already a logo component available in `@/components/app`, so you should be able to use that. You may need to add a wrapper component to get it centered above the `Empty` and with sufficient space.
2. Can we have two buttons here? One for "Go back", which uses `history` from TanStack router to go back to the last page (see [this comment](https://github.com/TanStack/router/discussions/181#discussioncomment-8407749) for usage). The second "button" would actually be passed in as a `link` prop and would take the user to the home index route, like you're doing now. It could say "Go home".

Let me know if this is clear or if you need more support. Thanks again for contributing!

.changeset/weak-numbers-stare.md
```diff
@@ -0,0 +1,5 @@
+---
+"namesake": minor
+---
+
+Improved the 404 page by adding navigation to the home page, enhancing user experience and accessibility.
```
kydecker:
```suggestion
Add navigation links and logo to the 404 page
```
Generally I try to keep these changelog messages short!

src/main.tsx
```diff
@@ -31,6 +31,14 @@ const router = createRouter({
       title="Page not found"
       subtitle="We couldn't find that page."
       icon={TriangleAlert}
+      button={{
+        children: "Go back to home page",
+        onPress: () => {
+          router.navigate({ to: "/" });
+        },
+        variant: "secondary",
+        size: "medium",
```
kydecker:
```suggestion
```
Since the variant defaults to `secondary` and the size defaults to `medium`, we can exclude both of these.

convex/_generated/api.d.ts
```diff
@@ -26,8 +26,8 @@ import type * as seed from "../seed.js";
 import type * as topics from "../topics.js";
 import type * as userEncryptedData from "../userEncryptedData.js";
 import type * as userQuests from "../userQuests.js";
-import type * as userSettings from "../userSettings.js";
 import type * as users from "../users.js";
+import type * as userSettings from "../userSettings.js";
```
kydecker: Looks like this may have gotten committed by accident.

.gitignore
```diff
@@ -40,3 +40,5 @@ node_modules
 
 # Local files
 *.local
+
+.env.local
```
kydecker:
```suggestion
.env.local

# Local files
*.local

```

kydecker:
This is great! Thank you for your contribution! I made a few small design tweaks to the component structure so that we can put the buttons on the same row.

CleanShot 2024-12-04 at 11 44 05@2x

kydecker (approved):
LGTM!

PR #268, opened 2024-12-07
feat: Adjust user role assignment based on environment

### LINKED ISSUE #260: Default user role to "admin" during signup if environment is dev
During signup, we assign all new users the “user” role. It would be good to default to “admin” if we detect the environment is development instead of production. This will assist new developers with accessing admin-only settings and changing data from within the UI.

belhajManel:
## What changed?
Resolves #260 : The user role is now defaulted to "admin" during signup when the environment is set to "development."

## Why?
This change helps new developers by granting access to admin-only settings and allowing them to modify data directly from the UI.

## How was this change made?

- Added a conditional check during the signup process to assign the "admin" role if the environment is set to development.
- Kept the default role as “user” for production environments.

## How was this tested?
Manual testing by verifying admin-only features are accessible after signup in development mode.

## Anything else?
Documentation updates may be needed to inform developers about this behavior during local setup.

### DIFF
--- a/.changeset/funny-roses-wink.md
+++ b/.changeset/funny-roses-wink.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Default the user role to admin during signup when the environment is set to development.
--- a/convex/auth.test.ts
+++ b/convex/auth.test.ts
@@ -1,0 +1,57 @@
+import { faker } from "@faker-js/faker";
+import { convexTest } from "convex-test";
+import { describe, expect, it, vi } from "vitest";
+import { api } from "./_generated/api";
+import schema from "./schema";
+import { modules } from "./test.setup";
+
+describe("auth roles", () => {
+  it("should set role to admin in development environment", async () => {
+    const t = convexTest(schema, modules);
+
+    const dummyUser = {
+      email: faker.internet.email(),
+      emailVerified: faker.datatype.boolean(),
+    };
+
+    // Set NODE_ENV to development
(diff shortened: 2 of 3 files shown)


kydecker (commented):
🔥 On a roll with contributions! Thanks for adding this!

I'd love to add a unit test to verify that roles are set correctly depending on dev vs prod environments. Do you think you could add a new `auth.test.ts` file and write a basic test for this? You'll need to mock the dev environment using [`stubEnv`](https://vitest.dev/api/vi.html#vi-stubenv), and run tests with `pnpm test`.

belhajManel:
Thank you and sure, I’ll try adding the needed test case!

belhajManel:
I added 2 test cases to verify role assignment based on the environment. In the development environment, the role is set to "admin," while in production, it defaults to "user."
Is that what we need to cover?

kydecker:
Yes! Those are the test cases we need.

However, since we are redeclaring the conditional logic within the test file, if the `createOrUpdateUser` function in the app changes, the test will still pass, even if it should fail.

We need to test `createOrUpdateUser` directly, instead of redeclaring logic. It might require moving the function out of the callback and into its own function so that we can test it in Vitest. 

Let me know if you have any questions!

belhajManel:
That makes total sense for future maintainability.
I extracted `createOrUpdateUser` into a separate function and updated the test cases to use it.

Let me know if we need another tweak!

kydecker (approved):
Made a small formatting tweak, otherwise looks and works great! Thanks so much!

PR #303, opened 2024-12-21
fix: Truncate long text within buttons

### LINKED ISSUE #302: Truncate long text within buttons
We're using our `Button` component to display user data within the settings page and in the sidebar. Buttons should max out at a particular width by default (wide enough for the majority of real buttons without truncation, but not wide enough to cause problems for long user strings).

CleanShot 2024-12-20 at 23 15 35@2x

CleanShot 2024-12-20 at 23 15 43@2x

chuck-ankit:
## What changed?
The Button in the `EditNameSettings.tsx` component was updated to address the issue with overly long user names. 
- Added the `max-w-[200px]` class to limit the button width.
- Applied the `truncate` class to truncate long strings.
- Implemented a tooltip that appears on hover to show the full name for names longer than 15 characters.

Fixes #302 (Truncate long text within buttons)

## Why?
This change addresses the issue where long user names were overflowing within the button. It ensures that buttons maintain a consistent width and provides a way to view the full name using a tooltip on hover.

## How was this change made?
- The `max-w-[200px]` class was applied to restrict the button width.
- The `truncate` class was added to cut off overflowing text.
- A tooltip component was added that displays the full user name for names exceeding 15 characters when hovered over.

## How was this tested?
- Tested with different user names to verify that names longer than 15 characters are properly truncated.
- Hovered over long names to ensure the tooltip correctly displays the full name.
- Verified the layout and button resizing under different screen sizes to ensure responsiveness.

## Anything else?
- No further adjustments are needed unless additional features are requested. Let me know if anything should be modified.

### DIFF
--- a/.changeset/eight-yaks-itch.md
+++ b/.changeset/eight-yaks-itch.md
@@ -1,5 +1,0 @@
----
-"namesake": minor
----
-
-Added email display in account settings page.
--- a/.changeset/gorgeous-colts-reply.md
+++ b/.changeset/gorgeous-colts-reply.md
@@ -1,5 +1,0 @@
----
-"namesake": minor
----
-
-Add links to terms of service and privacy policy
--- a/.changeset/silent-cougars-beg.md
+++ b/.changeset/silent-cougars-beg.md
@@ -1,5 +1,0 @@
----
-"namesake": patch
----
-
-Ensure that accessible titles are available for all dialogs
--- a/.changeset/smart-flowers-retire.md
+++ b/.changeset/smart-flowers-retire.md
@@ -1,5 +1,0 @@
----
-"namesake": patch
----
-
-Ensure automatic tests run when upgrading packages
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,17 @@
 # namesake
 
+## 0.15.0
+
+### Minor Changes
+
+- 6e76b6f: Added email display in account settings page.
+- cf18a88: Add links to terms of service and privacy policy
+
+### Patch Changes
+
+- 3192f7e: Ensure that accessible titles are available for all dialogs
+- 2dcace8: Ensure automatic tests run when upgrading packages
+
 ## 0.14.0
 
 ### Minor Changes
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "namesake",
-  "version": "0.14.0",
+  "version": "0.15.0",
   "private": true,
   "type": "module",
   "scripts": {


.gitignore
```diff
@@ -42,3 +42,5 @@ node_modules
 
 # Local files
 *.local
+
+.env.local
```
kydecker: It looks like the Convex initialization added this automatically; can you revert it since `.env.local` is already included above?

convex/README.md
```diff
@@ -0,0 +1,90 @@
+# Welcome to your Convex functions directory!
```
kydecker: Convex initialization may have added this file, too. Mind removing it?

src/components/settings/EditNameSetting/EditNameSetting.tsx
```diff
@@ -99,9 +99,20 @@ export const EditNameSetting = ({ user }: EditNameSettingProps) => {
 
   return (
     <SettingsItem label="Name" description="How should Namesake refer to you?">
-      <Button icon={Pencil} onPress={() => setIsNameModalOpen(true)}>
-        {user?.name ?? "Set name"}
-      </Button>
+      <div className="relative group">
+        <Button
+          icon={Pencil}
+          className="max-w-[200px] truncate"
```
kydecker: We'll want to make this change on the shared `Button` component, instead of this _specific_ button. The `Button` component is in `src/components/common`!

src/components/settings/EditNameSetting/EditNameSetting.tsx
```diff
@@ -99,9 +99,20 @@ export const EditNameSetting = ({ user }: EditNameSettingProps) => {
...
+          onPress={() => setIsNameModalOpen(true)}
+        >
+          {user?.name ?? "Set name"}
+        </Button>
+        {user?.name && user.name.length > 15 && (
+          <span className="absolute hidden group-hover:block bg-gray-700 text-white text-sm px-2 py-1 rounded">
+            {user.name}
+          </span>
+        )}
```
kydecker: I like your idea of showing a tooltip when the text is too long! We have a `Tooltip` component (also in `src/components/common`) that we can use for that instead of a custom element.

Bonus: If we set `max-w-[30ch]` using the [`ch` unit](https://caniuse.com/ch-unit), we can reliably show the `Tooltip` when the contents are more than 30 characters long. This will also make sure if we increase the font size on `Button`, the same number of characters will be displayed!

kydecker (commented):
Hey @chuck-ankit, thanks for your contribution!

kydecker:
Note: we should also add a changeset to this PR! You can add one by running `pnpm changeset`, selecting "patch", and entering a comment from the user's point of view like "Truncate long text within buttons"

PR #308, opened 2024-12-22
feat: integrate `zxcvbn-ts` for password strength evaluation

### LINKED ISSUE #287: Add password strength requirements during registration and password reset
We should implement password requirements to reduce the risk of phishing and account loss.

https://labs.convex.dev/auth/config/passwords#customize-sign-up-form-validation

belhajManel:
## What changed?
Resolves #287 
Add password strength evaluation during registration and password reset.

https://github.com/user-attachments/assets/626ed7e3-5bcf-42f0-a2e4-ffee98523202

## Why?
Limit account phishing by requiring a minimum password score when creating an account or reseting a password.

## How was this change made?

- Integrated `zxcvbn-ts` for password strength estimation
- Created a custom `usePasswordStrength` hook to use in different components 

## How was this tested?
👀


src/components/app/PasswordStrength/PasswordStrength.tsx
```diff
@@ -140,17 +140,17 @@ export function PasswordStrength({
         )}
       </AriaMeter>
       {warning && <Banner variant="danger">{warning}</Banner>}
-      {suggestions && (
+      {suggestions && suggestions.length > 0 && (
         <Banner variant="info">
-          <p>
+          <span>
```
belhajManel: I made this tiny change because it will cause a hydration error

src/utils/usePasswordStrength.ts
```diff
@@ -0,0 +1,36 @@
...
+    },
+    translations: zxcvbnEnPackage.translations,
+  };
+};
+
+loadOptions().then((options) => {
+  zxcvbnOptions.setOptions(options);
+});
+
+export function usePasswordStrength(password: string) {
```
kydecker, line 25: Love the creation of a hook for this!

src/utils/usePasswordStrength.ts
```diff
@@ -0,0 +1,36 @@
...
+    /* webpackChunkName: "zxcvbnEnPackage" */ "@zxcvbn-ts/language-en"
+  );
+  return {
+    dictionary: {
+      ...zxcvbnCommonPackage.dictionary,
+      ...zxcvbnEnPackage.dictionary,
+    },
+    translations: zxcvbnEnPackage.translations,
+  };
+};
```
kydecker, line 19: Thanks for adding lazy-loading!

src/routes/_unauthenticated/signin.tsx
```diff
@@ -278,7 +295,18 @@ const ForgotPassword = ({
...
+        <PasswordStrength
+          value={passwordState.score}
+          warning={passwordState.feedback.warning || undefined}
+          suggestions={passwordState.feedback.suggestions || undefined}
+        />
+      )}
+      <Button
+        type="submit"
+        variant="primary"
+        isDisabled={isSubmitting || (passwordState && passwordState?.score < 3)}
```
kydecker: Instead of disabling the button entirely, which can cause [accessibility issues](https://axesslab.com/disabled-buttons-suck/), can we instead update our form submission handler to set the `error` when the score is `< 3`?

It might also be nice to pull the submission handler out into a `handleSubmit` function for readability.

src/routes/_unauthenticated/signin.tsx
```diff
@@ -137,7 +140,20 @@ const SignIn = () => {
...
+                value={passwordState.score}
+                warning={passwordState.feedback.warning ?? undefined}
+                suggestions={passwordState.feedback.suggestions ?? undefined}
+              />
+            )}
+            <Button
+              type="submit"
+              isDisabled={
+                isSubmitting || (passwordState && passwordState.score < 3)
+              }
```
kydecker: Let's display an error instead of disabling the button (see other comment)

kydecker (commented):
Yay! This rocks. 🔥🔥🔥

First: Can we tighten up the margin here between the input and strength meter?

CleanShot 2024-12-22 at 10 44 18@2x

Second: seeing this in action, the messages appearing and disappearing while typing is distracting. I wonder if we can update the `PasswordStrength` component to:

- Display a [`circle-help` icon](https://lucide.dev/icons/circle-help) to the right of the label when suggestions or a warning exists
- Open a `Popover` with the existing banners
- `Popover` should display on the right when there's available space, otherwise on the bottom

CleanShot 2024-12-22 at 10 52 11@2x

Third: Can we update the `PasswordStrength` component to prevent the label from re-animating if the strength does not change?

https://github.com/user-attachments/assets/8c137b99-b7e6-4b0d-a521-0cca23be7183

Let me know if you have any thoughts/questions/recommended improvements for the design or implementation!

src/routes/_unauthenticated/signin.tsx
```diff
@@ -137,7 +140,20 @@ const SignIn = () => {
               value={password}
               onChange={setPassword}
             />
-            <Button type="submit" isDisabled={isSubmitting} variant="primary">
+            {password && passwordState && (
```
kydecker, line 188: One more idea:

1. Can we change the api of `PasswordStrength` to just accept one prop for `password`, instead of `value` and `warning` and `suggestions`? And then handle all logic for display inside of the component? That way we don't have to check `?? undefined` or nest it within a conditional statement wherever it's used, and we can invoke the hook from within the component itself.
2. Can we display the strength meter *even when* the password is empty, but hide the label? To prevent reflowing the page when a user begins typing.

belhajManel: Definitely! 
For the 2nd point when the password is empty do we provide the help icon or is it hidden with the label? It can have initial suggestions for the pwd structure but idk if that really helps?

belhajManel:
The `Popover` idea is neat⚡ I'll make the adjustments.

src/routes/_unauthenticated/signin.tsx
kydecker: Does the library return suggestions if the password is empty?

belhajManel: Yep it does return 2 suggestions

kydecker: Yeah, let's try leaving it in even without a label.

kydecker:
@belhajManel Let me know if you need any help moving this forward! Happy to contribute to the popover / design updates.

kydecker:
@belhajManel Unfortunately I need to close this PR. Namesake is switching to Clerk for authentication, which provides its own password strength checking.

Thank you for your work here!

kydecker:
Re-opening 😅😅

Clerk didn't feel like a good fit, so we're going to stick with Convex Auth for the time being. I can try to get this across the finish line!

kydecker:
Made a few changes here to simplify the design and implementation!

- Only animate the strength label when the value changes
- Eliminate the `setInterval()` timeout for showing and hiding characters one-by-one, which posed issues when rapidly changing between values
- Eliminate the suggestion banners, for now
- Put the warning message, if one exists, in the error banner at the top, triggered on submit
- Do not disable the submit/registration buttons, but display an error if the password is not strong enough

https://github.com/user-attachments/assets/3c4dd087-02aa-4832-b4af-e6d0a6c52fdb

Thank you so much @belhajManel for all your work on this!

kydecker (approved):

PR #365, opened 2025-02-01
docs: Add Git LFS installation note

fartinmartin:
## What changed?
I've just set up my local environment by following [CONTRIBUTING.md](/docs/CONTRIBUTING.md#getting-started) and ran into a few hiccups:

1. When cloning the repo the initial `checkout` failed as I did not have `git-lfs` installed, so I added a note directing folks to install instructions.
2. I wasn't sure if there is a way to login to your local Namesake app without an account, so I added instructions for how I created an early access code and registered an account.

PR #368, opened 2025-02-02
fix: Remove early access code requirement in dev

### LINKED ISSUE #366: Allow local registration without early access code
#352 introduced early access codes to limit production access, but it also removed a check for `import.meta.env.PROD`.

**Problem:**
When registering locally, an early access code should not be required.

**Fix:**
Add a check for `import.meta.env.PROD` to decide whether to display the early access code requirement. If a user is in a development environment, allow them to register without an early access code.

fartinmartin:
## What changed?
Fixes #366

## Why?
Easier dev onboarding, see #365.


kydecker (approved):
Easy 😌

PR #379, opened 2025-02-07
docs: remove unused dep in CONTRIBUTING.md

atrakh:
## What changed?

Removed SurveyJS from `docs/CONTRIBUTING.md` as it's no longer in the dep list

Also updated `.gitignore` to ignore the generated `convex/README.md` file that gets created whenever you initialize a convex deployment


kydecker (approved):
Sweet! Thanks for the update 🏆

PR #380, opened 2025-02-07
feat: Encrypt user form data on client side

atrakh:
## What changed?

- Adds a set of encryption methods to encrypt user data before it is sent to the database (Convex)
- Adds a placeholder UI on the `/settings/data` page to help demonstrate how the feature works. The forms in the UI are basic and should be updated in the future to properly support intended use cases.
- Adds a `list` query to retrieve all `userFormData` for rendering

## Why?

Users of namesake should have there data protected -- external parties and operators of namesake should have a minimal amount of PII (Personally Identifiable Information) available to them

## How was this change made?

This PR implements a client-side encryption key, stored in IndexedDB in the browser, that is used to encrypt and decrypt data between the client and server.

I initially experimented with a two-key system (implementing a [Key Encrypting Key](https://csrc.nist.gov/glossary/term/key_encryption_key)), but found that implementation complicated and likely to not be completely secure. 

This simpler implementation is more streamlined, reducing the risk of vulnerability. However, there are some things to consider regarding security:
- To reduce the risk of XSS or other attacks where browser storage can be compromised, KEK could be implemented with a non-extractable key
- Other XSS prevention measure should exist in the application and be tested before 1.0

This implementation does not currently implement convenient portability of the key. If the user logs in from another device, a new key will be generated and data will not load. I recommend we merge this PR as a proof-of-concept and add the portability feature (e.g. allowing the user to download their key and re-enter it later) as a follow-up.

## How was this tested?

I added a basic test-suite to test round-trips -- encryption methods and IndexedDB are mocked. I also used the placeholder UI to test manually and confirm encrypted data is stored in Convex.

## Anything else?

Migrations! If, after launch, the key-generation / encryption patterns are changed in the future, we'll need to maintain the existing implementation and migrate the database to indicate which implementation was used to encrypt the data for backwards compatability

### DIFF
--- a/src/components/settings/UserDataForm/UserDataForm.tsx
+++ b/src/components/settings/UserDataForm/UserDataForm.tsx
@@ -80,6 +80,8 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
     return <LoaderCircle className="w-4 h-4 animate-spin" />;
   }
 
+  const isExistingField = initialData.field !== "";
+
   return (
     <Form onSubmit={handleSubmit}>
       <div className="flex gap-2 items-end">
@@ -89,6 +91,7 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
           value={field}
           onChange={setField}
           placeholder="Enter field name"
+          isDisabled={isExistingField}
         />
         <TextField
           label="Value"
--- a/src/components/settings/UserDataForm/UserDataForm.tsx
+++ b/src/components/settings/UserDataForm/UserDataForm.tsx
@@ -1,4 +1,4 @@
-import { Button, Form, TextField } from "@/components/common";
+import { Banner, Button, Form, TextField } from "@/components/common";
 import { decryptData, encryptData, getEncryptionKey } from "@/utils/encryption";
 import { api } from "@convex/_generated/api";
 import { useMutation } from "convex/react";
@@ -20,16 +20,14 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
   const [isSaving, setIsSaving] = useState(false);
   const isDirty =
     field === initialData.field && decryptedValue === initialDecryptedValue;
+  const [didError, setDidError] = useState(false);
 
   const save = useMutation(api.userFormData.set);
 
   useEffect(() => {
-    console.log("initialData", initialData);
     const loadEncryptionKey = async () => {
-      console.log("TRYING!")
       try {
         const key = await getEncryptionKey();
-        console.log("key", key);
         setEncryptionKey(key);
 
         if (!key) {
@@ -43,6 +41,7 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
             setInitialDecryptedValue(decryptedValue);
           } catch (error: any) {
             posthog.captureException(error);
+            setDidError(true);
           }
         } else {
           setDecryptedValue("");
@@ -50,6 +49,7 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
         }
       } catch (error: any) {
         posthog.captureException(error);
+        setDidError(true);
       }
     };
 
@@ -64,17 +64,24 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
       return;
     }
 
-    setIsSaving(true);
     try {
+      setIsSaving(true);
+
       // Encrypt the value before saving
       const encryptedValue = await encryptData(decryptedValue, encryptionKey);
       await save({ field, value: encryptedValue });
-    } catch (error) {
-      console.error("Error saving encrypted data:", error);
     } finally {
       setIsSaving(false);
     }
   };
+
+  if (didError) {
+    return (
+      <Banner variant="danger">
+        Error decrypting {initialData.field}
+      </Banner>
+    );
+  }
 
   if (encryptionKey === null || decryptedValue === undefined) {
     return <LoaderCircle className="w-4 h-4 animate-spin" />;
--- a/src/components/settings/UserDataForm/UserDataForm.tsx
+++ b/src/components/settings/UserDataForm/UserDataForm.tsx
@@ -24,9 +24,12 @@ export function UserDataForm({ initialData }: UserDataFormProps) {
   const save = useMutation(api.userFormData.set);
 
   useEffect(() => {
+    console.log("initialData", initialData);
     const loadEncryptionKey = async () => {
+      console.log("TRYING!")
       try {
         const key = await getEncryptionKey();
+        console.log("key", key);
         setEncryptionKey(key);
 
         if (!key) {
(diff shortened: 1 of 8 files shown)


kydecker:
This is great! Thank you so much for this extremely helpful contribution. It goes a long way toward getting Namesake ready for launch. And thank you for the detailed and thoughtful PR comment; it helps a lot in understanding what's going on and limitations to follow up on.

I'm filing a few issues here which address issues you mention in your PR comment:

- #381 
- #382 

Appreciate if you have any thoughts or color to add to either of those tickets! Feel free to comment with extra context or opinions.

For this PR:

1. Can you add a changeset to this PR which explains the nature of the changes, written in user-facing language? `pnpm changeset` → `minor` → enter explanation
2. The test failure is from Biome, and should be automatically fixable if you run `pnpm check:fix`
3. If the Data Encryption Key is deleted manually, or unavailable, we get infinite loaders. Could we present a basic user-facing error banner if decryption fails?

https://github.com/user-attachments/assets/7b5acc97-5371-496f-8ae5-b291e180f313

src/components/settings/UserDataForm/UserDataForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+interface UserDataFormProps {
+  initialData: { field: string; value: string };
+}
+
+// Placeholder for allowing users to view and edit their form data
+// TODO: Replace this with a more robust implementation that manages draft state better.
+export function UserDataForm({ initialData }: UserDataFormProps) {
+  const [field, setField] = useState(initialData.field);
+  const [decryptedValue, setDecryptedValue] = useState<string>();
+  const [initialDecryptedValue, setInitialDecryptedValue] = useState<string>();
```
kydecker, line 18:
```suggestion
  const [decryptedValue, setDecryptedValue] = useState('');
  const [initialDecryptedValue, setInitialDecryptedValue] = useState('');
```
This is a matter of preference but for strings I generally like to have TypeScript infer the type. 

Should `initialDecryptedValue` be initialized with `initialData.value`? (Maybe not, since it will flash the encrypted value before `useEffect` runs?)

src/components/settings/UserDataForm/UserDataForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  const save = useMutation(api.userFormData.set);
+
+  useEffect(() => {
+    console.log("initialData", initialData);
+    const loadEncryptionKey = async () => {
+      console.log("TRYING!")
+      try {
+        const key = await getEncryptionKey();
+        console.log("key", key);
```
kydecker: We'll want to remove these `console.log` statements

src/components/settings/UserDataForm/UserDataForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+          return;
+        }
+
+        if (initialData.value) {
+          try {
+            const decryptedValue = await decryptData(initialData.value, key);
+            setDecryptedValue(decryptedValue);
+            setInitialDecryptedValue(decryptedValue);
+          } catch (error: any) {
+            posthog.captureException(error);
```
kydecker, line 43: Thanks for adding `posthog` in here! I actually have yet to implement `captureException` and other posthog events, so this is a good place to start. 

Could you add `captureException` to other `catch` statements, too?

src/components/settings/UserDataForm/UserDataForm.tsx
```diff
@@ -0,0 +1,123 @@
...
+  const handleSubmit = async (event: React.FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (!encryptionKey || !decryptedValue) {
+      console.error("No encryption key or decrypted value available");
+      return;
+    }
+
+    setIsSaving(true);
+    try {
```
kydecker:
```suggestion
    try {
      setIsSaving(true);
```
nit

src/components/settings/UserDataForms/UserDataForms.tsx
```diff
@@ -0,0 +1,56 @@
...
+  useEffect(() => {
+    const setupEncryption = async () => {
+      try {
+        await initializeEncryption();
+      } catch (error: any) {
+        posthog.captureException(error);
+      }
+    };
+
+    setupEncryption();
```
kydecker, line 31: What's the reason for having `setupEncryption` instead of calling `try / catch` directly in `useEffect`?

src/utils/encryption.test.ts
```diff
@@ -0,0 +1,234 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
```
kydecker: Thank you so much for writing all these mocks and tests!!

kydecker (commented):
A few small comments and questions, but otherwise very excited to get this merged!

src/components/settings/UserDataForm/UserDataForm.tsx
atrakh: whoops, i totally missed that these were still in here 😂

src/components/settings/UserDataForms/UserDataForms.tsx
atrakh: `initializeEncryption` is async so if it's not awaited, i don't think the catch will work. another option would be refactor all the error handling into `encryption.ts`

atrakh: I also wasn't sure what the ideal pattern would be for error handling -- do you want errors to be rethrown?

src/components/settings/UserDataForm/UserDataForm.tsx
atrakh: Yup, `initialDecryptedValue` is not set to `initialData.value` to ensure we never surface the encrypted value in UI. I also had the default value be `undefined` where `initialDecryptedValue === undefined` indicated that the async decryption process had not completed yet

src/components/settings/UserDataForm/UserDataForm.tsx
atrakh: Sure thing! however maybe we just want to let these errors throw and handle it in an error boundary? They're not really recoverable -- but may be more-so when we support portable encryption keys

atrakh:
Already! pushed up changes to address the review, apart from a couple comments where the explanation warrants keeping the code as is (IMO - let me know if you think we should still make changes)

Here's a screenshot of the basic error handling for failure to decrypt. This should be made better when portable keys are nted

Screenshot 2025-02-07 at 1 45 59 PM

src/components/settings/UserDataForm/UserDataForm.tsx
kydecker: Ah, got it! Makes sense!

kydecker (approved):
LGTM! Thanks so much for your contribution, @atrakh!

PR #387, opened 2025-02-09
fix: Wrap text correctly in `Disclosure` element

### LINKED ISSUE #378: Disclosure element does not wrap text correctly
image

Update the `Disclosure` component to allow the height to grow.

Text should remain left-aligned.

Caret icon should not shrink.

mbbischoff:
## What changed?

* Adjusted disclouse from height 10 to min-height 10
* Added `text-left` to align the button’s text to the left
* Added `shrink-0 ml-2` to the caret to prevent shrinking and make sure there’s always some space in the layout between the text and the disclousure arrow
* Top aligns controls with the question title

| Before | After  |
|--------|--------|
| Before Image |   CleanShot 2025-02-09 at 16 07 53@2x  |

Fixes https://github.com/namesakefyi/namesake/issues/378.

## Why?
This fixes layout bugs with longer questions.

## How was this change made?
This is my first commit on the project, so I wanted something very localized and not dangerous. I started with a `span` around the title, but then tried just adding `text-left` to the `disclosureTriggerStyles` and that seemed nicer.

## How was this tested?
I reproduced the bug with Lorem Ipsum text for long questions and answers, and tested the fix by scaling my browser text size up and down to make sure the layout remained legible.

### DIFF
--- a/src/components/common/Disclosure/Disclosure.tsx
+++ b/src/components/common/Disclosure/Disclosure.tsx
@@ -26,7 +26,7 @@ export interface DisclosureProps
 
 const disclosureTriggerStyles = tv({
   extend: focusRing,
-  base: "group flex-1 flex rounded-lg transition-colors text-gray-dim hover:text-gray-normal justify-between min-h-10 text-lg font-medium text-left",
+  base: "group flex-1 flex rounded-lg transition-colors text-gray-dim hover:text-gray-normal justify-between text-lg font-medium text-left py-2",
 });
 
 const disclosurePanelStyles = tv({
--- a/src/components/common/Disclosure/Disclosure.tsx
+++ b/src/components/common/Disclosure/Disclosure.tsx
@@ -56,8 +56,8 @@ export function Disclosure({
           <ChevronDown
             size={20}
             className={twMerge(
-              "group-hover:bg-graya-3 dark:group-hover:bg-graydarka-3 group-hover:text-gray-normal rounded-full size-8 p-1.5",
-              "transition-transform opacity-60 shrink-0 ml-2",
+              "group-hover:bg-graya-3 dark:group-hover:bg-graydarka-3 group-hover:text-gray-normal rounded-full size-8 p-1.5 shrink-0 ml-2",
+              "transition-transform opacity-60",
               "group-data-[expanded]:rotate-180",
             )}
           />
--- a/src/components/common/Disclosure/Disclosure.tsx
+++ b/src/components/common/Disclosure/Disclosure.tsx
@@ -26,7 +26,7 @@ export interface DisclosureProps
 
 const disclosureTriggerStyles = tv({
   extend: focusRing,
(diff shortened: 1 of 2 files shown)


kydecker:
This looks good to me! I wonder if it makes sense to top-align the caret and ellipsis?

mbbischoff:
> This looks good to me! I wonder if it makes sense to top-align the caret and ellipsis?

I was wondering that too, and that’s my preference, so let’s do it.

src/components/common/Disclosure/Disclosure.tsx
```diff
@@ -26,7 +26,7 @@ export interface DisclosureProps
 
 const disclosureTriggerStyles = tv({
   extend: focusRing,
-  base: "group flex-1 flex rounded-lg transition-colors text-gray-dim hover:text-gray-normal justify-between items-center h-10 text-lg font-medium",
```
mbbischoff, line 29: Should i be using `text-start` rather than `text-left` or is RTL not something we’re worried about at this point?

kydecker: Good q! I wouldn't worry about it for now. Unlikely to be supporting RTL languages any time soon.

kydecker:
Looking good!

Thoughts on design details:

https://github.com/user-attachments/assets/d175e2a5-79a8-42db-85a2-c69b1891cb39

Right now the single-line question appears off-center, and the multi-line question appears to have less space between the text and the answer content.

Maybe instead of enforcing `min-h-10`, we can set vertical padding with `py-1.5` or `py-2`? That way the padding will center the content. We might need to tweak the margins for the caret / ellipsis area so that it stays centered to the first line of text.

mbbischoff:
> Right now the single-line question appears off-center

Fixed this centering with padding rather than min-height as you suggested.

> and the multi-line question appears to have less space between the text and the answer content.

Little stuck on this one. Trying to figure out why that is and what the right fix would be.

> We might need to tweak the margins for the caret / ellipsis area so that it stays centered to the first line of text.

Took care of this.

kydecker:
Nice!

> Little stuck on this one. Trying to figure out why that is and what the right fix would be.

I can help: right now, the line height of the title text is less than the chevron button height. They both get aligned to the top, so on single-line questions it results in more space appearing between the question and answer content.

CleanShot 2025-02-09 at 15 46 29

We could fix this by:

1. Increasing the line height of the question to match the button height (but then our text leading becomes a little *too* spacious, so probably not the best approach)
2. Reducing the size of the chevron to match the existing line height (would need to see how it looks alongside the ellipsis button to make sure they look good together)
3. Adding top and bottom margins to the text content which is one half the size of the extra space (would require adding an extra `span` around the title text)

Out of these options I'm tempted to go with the 3rd one!

mbbischoff:
> 3\. Adding top and bottom margins to the text content which is one half the size of the extra space (would require adding an extra `span` around the title text)

CleanShot 2025-02-09 at 16 06 10@2x

Good call. I don’t mind that at all.

mbbischoff:
Updated screenshot and marking as ready for review

kydecker (approved):
Looks great! First contribution DOWN 😄

Issue #71, opened 2024-09-18
Set up tables to handle migrations


evadecker:
https://stack.convex.dev/intro-to-migrations
https://stack.convex.dev/migrating-data-with-mutations
https://www.npmjs.com/package/convex-helpers#stateful-migrations

evadecker:
Not necessary now; closing until relevant.

evadecker:
I love closing issues only to open them 5 days later

mbbischoff:
I love closing it 2 days later again with a PR.

PR #388, opened 2025-02-09
feat: Add last updated date to quests

### LINKED ISSUE #383: Display "last updated" date and time on all quests
Each quest page should display a "last updated" date and time somewhere on the page. The last updated timestamp should be saved whenever someone edits quest content or quest steps.

This will help provide transparency around freshness of information.

mbbischoff:
## What changed?

Added a **Last Updated** field to Quests.

CleanShot 2025-02-11 at 16 18 52
CleanShot 2025-02-09 at 18 05 43@2x

Fixes https://github.com/namesakefyi/namesake/issues/383.

## Why?
It’s good to know how fresh the information is in a quest, both for users and for admins.

## How was this change made?

- [x] Add `updatedAt` to the schema for Quests
- [x] Set `updatedAt` every time a quest is edited
- [x] Add `QuestUpdated` component that shows when a quest was last updated
- [x] Add Updated column to admin panel and shorten date format for easier reading
- [x] Make simple quest updates go through `updateQuest`
- [x] Added relative date formatting
- [x]  Tracked which user updated a quest.
- [x] Tested that updating steps and FAQs marked the quest as updated.
- [x] Added tests

There wasn’t a design specifed, so I went with adding a `StatGroup` for now in order to get feedback. Is this too prominent or just right?

## How was this tested?

* Open the Admin screen
* Edit a quest
* Make sure the updated date changes (did this when i was showing a longer date format)
* Confirm that old quests with non-existent updatedAt’s have a good default ("Never")
* Test editing Quests to make sure I didn’t break that with my refactor
* Ran tests

## Anything else?
- [x] I want to add tests here, but I figured I should get a read on the general approach/design first, just so I don’t test a bunch of stuff I then end up changing immediately.
- [x] Might be nice to use a `<time>` element, but `StatGroup` can only take a string at the moment, so I wasn’t sure if that refactor would be a good use of time.
- [x] **Should we also track the same user for initial creation or leave it blank?**


convex/quests.ts
```diff
@@ -146,50 +147,59 @@ export const create = userMutation({
...
+function updateQuest(
+  ctx: Parameters<typeof userMutation.call>[1],
+  questId: Id<"quests">,
+  update: Partial<Doc<"quests">>,
+) {
+  return ctx.db.patch(questId, {
+    ...update,
+    updatedAt: Date.now(),
+  });
+}
```
kydecker: I like this helper! We should be able to use the `MutationCtx` type here, see [Splitting up mutation code via helpers](https://docs.convex.dev/functions/mutation-functions#splitting-up-mutation-code-via-helpers)

convex/quests.ts
```diff
@@ -204,6 +214,7 @@ export const deleteUrl = userMutation({
 
     await ctx.db.patch(args.questId, {
       urls: existingUrls.filter((u) => u !== args.url),
+      updatedAt: Date.now(),
```
kydecker: Why not use the helper function here?

convex/quests.ts
```diff
@@ -265,6 +274,7 @@ export const addStep = userMutation({
 
     await ctx.db.patch(args.questId, {
       steps: [...existingSteps, newStep],
+      updatedAt: Date.now(),
```
kydecker: Ditto

convex/quests.ts
```diff
@@ -287,6 +297,7 @@ export const deleteStep = userMutation({
 
     await ctx.db.patch(args.questId, {
       steps: updatedSteps,
+      updatedAt: Date.now(),
```
kydecker: Ditto

convex/quests.ts
```diff
@@ -308,7 +319,10 @@ export const addFaq = userMutation({
 
     const existingFaqs = quest.faqs || [];
 
-    await ctx.db.patch(args.questId, { faqs: [...existingFaqs, questFaqId] });
+    await ctx.db.patch(args.questId, {
+      faqs: [...existingFaqs, questFaqId],
+      updatedAt: Date.now(),
+    });
```
kydecker: Ditto

convex/quests.ts
```diff
@@ -321,7 +335,10 @@ export const deleteFaq = userMutation({
 
     const updatedFaqs = quest.faqs?.filter((id) => id !== args.questFaqId);
 
-    await ctx.db.patch(args.questId, { faqs: updatedFaqs });
+    await ctx.db.patch(args.questId, {
+      faqs: updatedFaqs,
+      updatedAt: Date.now(),
+    });
```
kydecker: Ditto

src/components/quests/QuestUpdated/QuestUpdated.tsx
```diff
@@ -0,0 +1,18 @@
+import { StatGroup } from "@/components/quests";
```
kydecker: Oh, nice, I hadn't thought of putting this in the stats at the top! Maybe too prominent, though.

What do you think of just a small line of gray text at the bottom of the page?

"Last updated on `<updatedAt>`"

This will also allow us to use the `time` element.

(Related: what are your thoughts on relative dates like "3 days ago" with a tooltip for the full time?)

kydecker (commented):
Nice!

We need to make sure that edits to `questSteps` and `questFaqs` propagate up to `updatedAt` for `quests`.

Left some other feedback!

kydecker:
Don't want to add too much scope creep here, but what do you think about also adding an `updatedBy` field to store a `userId`? Even if we don't publicly display user info, seems handy to store.

mbbischoff:
> Don't want to add too much scope creep here, but what do you think about also adding an `updatedBy` field to store a `userId`? Even if we don't publicly display user info, seems handy to store.

Seems like a good idea. I’ll look at this and the comments.

src/components/quests/QuestUpdated/QuestUpdated.tsx
mbbischoff: Pushed moving it to the footer and moving it to relative date becuase I also kinda wanted that. Find it sort of weird that the tooltip seems to require a button because clicking this button doesn’t do anything.

mbbischoff:
@evadecker can you take one more look here at the design and implementation and if everything looks good I’ll figure out what we wanna test.

src/components/TimeAgo.tsx
```diff
@@ -0,0 +1,9 @@
+import RelativeTime from "react-relative-time";
```
kydecker: I like having a component for this. We don't need a new dependency, though! The built-in `Intl.RelativeTimeFormat` should handle it

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/RelativeTimeFormat

And then we'll want to use the same format from the package, e.g. `<time dateTime={machineReadable}>{humanReadable}</time>`

src/components/quests/QuestPageFooter/QuestPageFooter.tsx
```diff
@@ -0,0 +1,14 @@
+import type { Doc } from "@convex/_generated/dataModel";
+import { QuestUpdated } from "../QuestUpdated/QuestUpdated";
+
+interface QuestPageFooterProps {
+  quest: Doc<"quests">;
+}
+
+export function QuestPageFooter({ quest }: QuestPageFooterProps) {
+  return (
+    <div className="border-t border-gray-dim py-4">
```
kydecker:
```suggestion
    <footer className="border-t border-gray-dim py-4">
```
Lil semantic improvement

src/components/quests/QuestUpdated/QuestUpdated.tsx
```diff
@@ -0,0 +1,28 @@
...
+      <Button variant="ghost" className="text-gray-dim text-sm">
+        Updated {updatedTime}
+      </Button>
+      <Tooltip>
+        {quest.updatedAt
+          ? new Date(quest.updatedAt).toLocaleString()
+          : "Never updated"}
+      </Tooltip>
+    </TooltipTrigger>
+  );
```
kydecker: You ran into an intentional limitation with React Aria which is that tooltip triggers must be hoverable/focusable. In this case it feels a little odd to have a button, so I'm okay using the browser-default `title` attribute for the full time instead

src/routes/_authenticated/admin/quests.index.tsx
```diff
@@ -174,7 +175,16 @@ const QuestTableRow = ({
...
+      <TableCell>
+        <TimeAgo date={new Date(quest._creationTime)} />
+      </TableCell>
+      <TableCell>
+        {quest.updatedAt ? (
+          <TimeAgo date={new Date(quest.updatedAt)} />
+        ) : (
+          "Never"
+        )}
+      </TableCell>
```
kydecker: I actually like using the full date within tables. Using relative dates makes it hard to compare between rows where the units are different, whereas `MM/DD/YYYY` is easy to scan.

Edit: `MM/DD` can be ambiguous across so maybe better to have month shortnames, e.g. "Jan 15, 2024"

kydecker (commented):
Some design thoughts and an opportunity to rely on built-in `Intl.RelativeTimeFormat` instead of adding a dependency

src/components/quests/QuestUpdated/QuestUpdated.tsx
kydecker: Addressed in review!

convex/quests.test.ts
```diff
@@ -108,6 +108,8 @@ describe("quests", () => {
       expect(quest?.jurisdiction).toBe("MA");
       expect(quest?.slug).toBe(slug);
       expect(quest?.creationUser).toBe(userId);
+      expect(quest?.updatedAt).toBeCloseTo(Date.now(), -3);
```
kydecker: TIL about `toBeCloseTo`! `-3` is to set the precision?

src/components/FormattedDate.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const FormattedDate = ({ date }: FormattedDateProps) => (
+  <time dateTime={date.toISOString()}>
+    {date.toLocaleDateString(navigator.language, {
+      month: "short",
+      day: "numeric",
+      year: "numeric",
+    })}
+  </time>
+);
```
kydecker, line 13: Nice!

src/components/FormattedDate.test.tsx
```diff
@@ -0,0 +1,44 @@
...
+    expect(timeElement).toHaveAttribute("datetime", "2024-03-14T12:00:00.000Z");
+    expect(timeElement.textContent).toBe("Mar 14, 2024");
+  });
+
+  it("renders date in UK format when locale is en-GB", () => {
+    // Mock navigator.language
+    Object.defineProperty(navigator, "language", {
+      value: "en-GB",
+      configurable: true,
+    });
```
kydecker, line 35: Testing internationalization, hell yeah

src/components/TimeAgo.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+    ];
+
+    for (const { date, expected } of testCases) {
+      expect(getRelativeTimeString(date, "en")).toBe(expected);
+    }
+  });
+
+  it("formats times in the future correctly", () => {
+    const now = new Date("2024-03-20T12:00:00Z");
+    vi.setSystemTime(now);
```
kydecker, line 64: TIL about `setSystemTime`!

src/components/quests/QuestUpdated/QuestUpdated.tsx
```diff
@@ -0,0 +1,15 @@
...
+import type { Doc } from "@convex/_generated/dataModel";
+
+interface QuestUpdatedProps {
+  quest: Doc<"quests">;
+}
+export function QuestUpdated({ quest }: QuestUpdatedProps) {
+  const updatedTime = quest.updatedAt ? (
+    <TimeAgo date={new Date(quest.updatedAt)} />
+  ) : (
+    "Never"
```
kydecker: CleanShot 2025-02-11 at 20 00 59@2x

This feels strange 🤔 

```suggestion
    "some time ago"
```

If the user is viewing the quests, it exists, so it's been updated at *some* point. Ideally they never see this state, but maybe a gradual degradation could be to say, "Updated some time ago" as a fallback?

Edit: if we make `updatedAt` a required field in the schema it should avoid this entirely.

convex/schema.ts
```diff
@@ -56,11 +56,17 @@ const quests = defineTable({
   steps: v.optional(v.array(v.id("questSteps"))),
   /** Questions related to the quest. */
   faqs: v.optional(v.array(v.id("questFaqs"))),
+  /** Time in ms since epoch when the quest was last updated. */
+  updatedAt: v.optional(v.number()),
```
kydecker, line 60:
```suggestion
  updatedAt: v.number(),
```

I wonder if we just make this required in order to avoid the fallback "never" state?

src/routes/_authenticated/admin/quests.index.tsx
```diff
@@ -174,7 +175,16 @@ const QuestTableRow = ({
...
-      <TableCell>{new Date(quest._creationTime).toLocaleString()}</TableCell>
+      <TableCell>
+        <FormattedDate date={new Date(quest._creationTime)} />
+      </TableCell>
+      <TableCell>
+        {quest.updatedAt ? (
+          <FormattedDate date={new Date(quest.updatedAt)} />
+        ) : (
+          "Never"
+        )}
```
kydecker: If we make `updatedAt` required it should avoid this conditional.

kydecker (commented):
This is so comprehensive! Thank you for the thoughtful new components to handle dates (and all their test cases)!

We'll also need to update the mutations in `questSteps.ts` and `questFaqs.ts` to pass the `updatedBy` and `updatedAt` back up to the related `quest`.

convex/schema.ts
kydecker: Followup to "just make it required":

- You'll get a Convex error if you try to make the change because it enforces the schema against the table in your dev environment.
- When this happens I typically just delete the dev data and the schema validates. It's easy to do this from the Convex dashboard.
- For prod, however, we would need to create a migration and backfill data. I haven't actually had to set up migrations yet, since there's been so little data in prod. But now might be a good time. There's a [Migrations Convex component](https://www.convex.dev/components/migrations) which we should set up and get comfortable using before release. This [Intro to Migrations](https://stack.convex.dev/intro-to-migrations) guide from Convex is also helpful. Related issue: #71 

Writing all of this out now... suggested order of operations:

1. Get this merged **as-is** with the field optional
2. Set up migrations using the Convex component
3. Write a migration to backfill the `updatedAt` data for quests in prod
4. Make the `updatedAt` field required and remove conditionals in the frontend

mbbischoff: Filed here for myself, but maybe it’d be good to add me to the org at some point so I can assign / put in the milestone, etc. https://github.com/namesakefyi/namesake/issues/393

mbbischoff:
>  Should we also track the same user for initial creation or leave it blank?

So just to clarify, lock in this decision, we’re gonna treat "creation" as an update and set the `updatedAt`/`updatedBy` on creation. I’ll take care of that on my lunch break

src/components/FormattedDate.tsx
mbbischoff: thanks! eventually we can make the formats arguments if we want different date formats in different contexts :)

src/components/TimeAgo.test.tsx
mbbischoff: i love manipulating time and space

src/routes/_authenticated/admin/quests.index.tsx
```diff
@@ -174,7 +175,16 @@ const QuestTableRow = ({
...
       <TableCell>{questCount}</TableCell>
-      <TableCell>{new Date(quest._creationTime).toLocaleString()}</TableCell>
+      <TableCell>
+        <FormattedDate date={new Date(quest._creationTime)} />
+      </TableCell>
+      <TableCell>
+        {quest.updatedAt ? (
+          <FormattedDate date={new Date(quest.updatedAt)} />
+        ) : (
+          "Never"
```
mbbischoff:
```suggestion
          "some time ago"
```

convex/quests.test.ts
mbbischoff: Found an even better way, faking the system type and then we don’t have to round!

mbbischoff:
> This is so comprehensive! Thank you for the thoughtful new components to handle dates (and all their test cases)!
> 
> We'll also need to update the mutations in `questSteps.ts` and `questFaqs.ts` to pass the `updatedBy` and `updatedAt` back up to the related `quest`.

Taken care of!

kydecker (approved):
Looks great! Nice work on the tests with the timestamps, I like the pattern you landed on.

mbbischoff:
> Looks great! Nice work on the tests with the timestamps, I like the pattern you landed on.

thanks! snuck in a couple of key dates (and months where we don't know the date) as the test values case why not

kydecker:
Omg sick I didn't even spot that 👀

Good to merge whenever 🫡

PR #392, opened 2025-02-12
fix: Show red border when Field is invalid

### LINKED ISSUE #244: Red border not showing for error/invalid fields
**Describe the bug**
When a field has an error/invalid state, its border should be red.

**To Reproduce**
1. Go to any form with required fields (like sign in)
2. Attempt to submit the form without data
3. View the error state for the field
4. No red border is displayed

**Expected behavior**
A red border should display.

**Screenshots**
CleanShot 2024-11-29 at 22 06 36@2x

 - OS: macOS
 - Browser: Chrome

**Additional context**
Invalid styles are set for fields, but something about the current composition of components in React Aria is preventing those styles from applying. Figure out what is wrong.

fartinmartin:
## What changed?
Invalid fields will now be styled with the red border defined in `fieldBorderStyles.variants.isInvalid`. This should fix #244.

## Why?
By default, a React Aria `<Group>` component doesn't receive the context/render props of it's parent Field component. Because of this, our `fieldBorderStyles.variants.isInvalid` classes were never applied. To ensure the `<Group>` receives the render props we have to use a [render prop function](https://react-spectrum.adobe.com/react-aria/styling.html#render-props) at the root of the parent:

```tsx
<TextField>
  {(renderProps) => (
    <>
      <Group {...renderProps}>
        {/* ... */}
      </Group>
    </>
  )}
</TextField>
```

## How was this change made?
Each Field component (`TextField`, `SearchField`, etc) that uses a `<FieldGroup>` has been updated to use a render prop function that passes `{...renderProps}` to the `<FieldGroup>`.

## How was this tested?
Mostly by spamming `/signin`, but also clicking around after running `pnpm storybook`.

## Anything else?
Perhaps not all of these Field components need this? What's the best way to test/determine this?

### DIFF
--- a/.changeset/pretty-balloons-yell.md
+++ b/.changeset/pretty-balloons-yell.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Fix error borders for invalid fields
--- a/src/components/common/ComboBox/ComboBox.tsx
+++ b/src/components/common/ComboBox/ComboBox.tsx
@@ -45,25 +45,29 @@ export function ComboBox<T extends object>({
         "group flex flex-col gap-1",
       )}
     >
-      <Label>{label}</Label>
-      <FieldGroup>
-        <Input />
-        <Button
-          variant="icon"
-          className="w-7 h-7 p-0 mr-1 outline-offset-0"
-          icon={ChevronDown}
-        />
-      </FieldGroup>
-      {description && <FieldDescription>{description}</FieldDescription>}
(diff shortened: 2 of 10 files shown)


kydecker (approved):
Ah! Nice find and fix! This seems like the idiomatic React Aria way to do it.

CleanShot 2025-02-11 at 20 20 44@2x

> Perhaps not all of these Field components need this? What's the best way to test/determine this?

Seems fine adding it to all of them.

PR #397, opened 2025-02-13
fix: `RichText` should not display focus ring when non-editable

### LINKED ISSUE #353: `RichText` displays focus ring when non-editable
Currently, the `RichText` component displays a focus ring even when the component is display-only. Fix this to prevent the focus styles from displaying unless `RichText` is `editable`.

Image

kydecker:
## What changed?
Do not display a focus ring on `RichText` component when the component is not editable. Resolves #353 

| Before | After |
|--------|--------|
| CleanShot 2025-02-13 at 17 16 38@2x | CleanShot 2025-02-13 at 17 15 51@2x | 

## Why?
A focus ring should only display for the text area when it's in an editable state.

## How was this change made?
Apply `focus-within` styles.

## How was this tested?
Visual testing.

### DIFF
--- a/.changeset/wild-jeans-raise.md
+++ b/.changeset/wild-jeans-raise.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Fix a bug where a focus ring would display on non-editable rich text fields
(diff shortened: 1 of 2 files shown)


mbbischoff (approved):
Super simple, love it.

Issue #393, opened 2025-02-12
Make Quests `updatedAt` required and migrate.


mbbischoff:
Followup to "just make it required":

- You'll get a Convex error if you try to make the change because it enforces the schema against the table in your dev environment.
- When this happens I typically just delete the dev data and the schema validates. It's easy to do this from the Convex dashboard.
- For prod, however, we would need to create a migration and backfill data. I haven't actually had to set up migrations yet, since there's been so little data in prod. But now might be a good time. There's a [Migrations Convex component](https://www.convex.dev/components/migrations) which we should set up and get comfortable using before release. This [Intro to Migrations](https://stack.convex.dev/intro-to-migrations) guide from Convex is also helpful. Related issue: #71 

Writing all of this out now... suggested order of operations:

1. Get this merged **as-is** with the field optional
2. Set up migrations using the Convex component
3. Write a migration to backfill the `updatedAt` data for quests in prod
4. Make the `updatedAt` field required and remove conditionals in the frontend

_Originally posted by @evadecker in https://github.com/namesakefyi/namesake/pull/388#discussion_r1951832023_

PR #398, opened 2025-02-14
chore: Create Migration for Quests.updatedAt in preparation for making it mandatory

mbbischoff:
## What changed?

- [x] Created a `setQuestsUpdatedTimestamp` migration that sets the `updatedAt` to now for all quests where it was undefined.

**Before:**
CleanShot 2025-02-14 at 15 38 14@2x
**During**:
CleanShot 2025-02-14 at 16 55 23@2x
**After**:
CleanShot 2025-02-14 at 16 55 49@2x

Progress on #393.

## Why?
So that we can run this in produciton and then make `quests.updatedAt` required.

## How was this change made?
* Followed the documentation for [**Defining Migrations**](https://www.convex.dev/components/migrations)

## How was this tested?
* Commented out setting updatedAt on new quests
* Made a new quest
* Ran the migration from the Dashboard in dev
* Confirmed that updatedAt was set
* Ran the new tests


mbbischoff:
Figured out a way to add a unit test for this, this afternoon.

.changeset/real-moons-wink.md
```diff
@@ -0,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Add a migration to allow us to make Quests require an updated time
```
kydecker: I haven't totally settled on how best to use changesets so far (open to suggestions), but right now I'm trying to mostly write them for user-facing changes, so that our changelog can be read and understood by end users.

In other words, I think something like this (setting up backend migrations) doesn't need a changeset. (It'll still get bundled in the release either way)

convex/migrations.test.ts
```diff
@@ -0,0 +1,81 @@
...
+import { internal } from "./_generated/api";
+import { migrations } from "./migrations";
+import schema from "./schema";
+import { modules } from "./test.setup";
+
+const migrationsModules = import.meta.glob(
+  "../node_modules/@convex-dev/migrations/src/component/**/*.ts",
+);
+
+// Tests inspired by https://github.com/get-convex/aggregate/blob/main/example/convex/aggregate.test.ts.
```
kydecker, line 13: Love tests for this! Nice!

What if we have a `/convex/migrations` folder where each file is a discrete migration?

E.g.
```
/convex
  /migrations
    setQuestsUpdatedTimestamp.ts
    setQuestsUpdatedTimestamp.test.ts
```

It also might be good to put together our own `migrations.md` guide for maintainers (or include a new section in `contributing.md`)? One that explains the steps to take, links to existing docs, etc. I assume we'll want to delete the migration files once the actual migration is complete, so we don't have tests and code lingering around in the codebase that are no longer useful.

.changeset/real-moons-wink.md
mbbischoff: Makes sense to me. I think the PR comment is a lil scary with the ⚠️ emoji, so maybe if that’s configurable we can make it a little softer so that folks know it’s only needed for user facing changes?

convex/migrations.test.ts
mbbischoff: Attempted to make this change but running into trouble with the imports and convex build process when I do. Also searched GitHub and docs to find an example of someone stucturing it that way and could only find https://github.com/Fuminshou-dev/DTrpg/blob/e80420bf25aaaf7b5b25b967fb0605334d4f976e/convex/migrations/add_brothel_status.ts#L4 which predates the official migrations component. Any ideas what I’m doing wrong here @evadecker?

.changeset/real-moons-wink.md
kydecker: Unfortunately it doesn't look like the PR comment body is currently customizable: https://github.com/changesets/action/issues/65

Not sure what else we could do 🤔 open to suggestions

convex/migrations.test.ts
kydecker: Hm, I'm not sure. Maybe worth asking in the [Convex Discord](https://discord.gg/ZVdQ4Fm9)?

If we can't refactor this to multiple files, it's not a huge deal, but I think separate files would make it easier to maintain, merge, and delete mutations as we go forward.

mbbischoff: I’m gonna ask there and move this to its own ticket so we don’t block on it

https://github.com/namesakefyi/namesake/issues/404

.changeset/real-moons-wink.md
mbbischoff: I think the sentence you’ve already got in the guidelines covers it well enough then. So I’m gonna resolve this one.

"PRs which only affect developers (refactoring some tests, for example) do not need a changeset."

convex/migrations.test.ts
mbbischoff: https://discord.com/channels/1019350475847499849/1341523547889205298/1341523547889205298

kydecker (approved):
Thanks for adding the followup tickets and keeping this PR moving! LGTM.

PR #400, opened 2025-02-17
feat: Display navigation controls within forms

### LINKED ISSUE #324: Add prev/next navigation to form sections
Each `FormSection` uses scroll snapping and appears as a full-height view.

https://github.com/user-attachments/assets/98661354-0374-4a8b-89b1-ae2ea87babc8

We should add prev/next navigation that allows users to jump between sections using a single click.

**Visual:**
Two caret icons (up/down) should display and float over the UI. Each button should navigate the user up or down one question.

kydecker:
## What changed?
Create a new `FormNavigation` component which displays a next button, previous button, and popover menu with all questions in a form. Resolves #324 

https://github.com/user-attachments/assets/c855ea0f-1716-4ea1-9f02-3c41f719b12b

## Why?
Make it easy for users to navigate through a form and jump to specific questions.

## How was this change made?
Created a new `IntersectionObserver` to monitor the currently-in-view question and update internal React state and the URL hash. Use that state to populate the menu and the URLs for the next and previous buttons.

## How was this tested?
- Added unit tests
- Manual testing

## Anything else?
Added some basic tests for `FormContainer`.

**Fix before merge:**
- [x] The "back" button on the form (labeled "save and exit") currently cycles through all of the updates to the hash in the URL. We should prevent hash changes from modifying the history of the page.

### DIFF
--- a/src/components/forms/FormNavigation/FormNavigation.test.tsx
+++ b/src/components/forms/FormNavigation/FormNavigation.test.tsx
@@ -1,0 +1,98 @@
+import { FormNavigation, FormSection } from "@/components/forms";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("FormNavigation", () => {
+  const mockSections = ["First Question", "Second Question", "Third Question"];
+
+  beforeEach(() => {
+    // Mock the DOM structure that FormNavigation expects
+    for (const title of mockSections) {
+      render(<FormSection title={title} />);
+    }
+
+    // Mock IntersectionObserver
+    const mockIntersectionObserver = vi.fn();
+    mockIntersectionObserver.mockReturnValue({
+      observe: () => null,
+      disconnect: () => null,
+    });
+    window.IntersectionObserver = mockIntersectionObserver;
+  });
+
+  afterEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it("renders navigation controls", () => {
+    render(<FormNavigation />);
+
+    expect(screen.getByLabelText("Previous question")).toBeInTheDocument();
+    expect(screen.getByLabelText("Next question")).toBeInTheDocument();
+    expect(
+      screen.getByRole("button", { name: "All questions" }),
+    ).toBeInTheDocument();
+  });
+
+  it("renders all questions in menu popover", async () => {
+    render(<FormNavigation />);
+
+    userEvent.click(screen.getByRole("button", { name: "All questions" }));
+
+    const menu = await screen.findByRole("menu");
+    expect(menu).toBeInTheDocument();
+
+    for (const title of mockSections) {
+      expect(screen.getByRole("menuitem", { name: title })).toBeInTheDocument();
+    }
+  });
+
+  it("disables previous button when at first section", async () => {
+    render(<FormNavigation />);
+
+    // Simulate first section being visible
+    const observer = window.IntersectionObserver.mock.calls[0][0];
+    observer([
+      {
+        target: document.getElementById("first-question"),
+        isIntersecting: true,
+      },
+    ]);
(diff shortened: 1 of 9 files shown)


src/components/forms/FormNavigation/FormNavigation.test.tsx
```diff
@@ -0,0 +1,98 @@
+import { FormNavigation, FormSection } from "@/components/forms";
+import { render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("FormNavigation", () => {
```
mbbischoff, line 6: Is it worth also testing the next/previous buttons do what we expect when activated?

src/components/forms/FormSection/FormSection.tsx
```diff
@@ -25,6 +28,16 @@ function FormHeader({ title, description }: FormHeaderProps) {
...
 
+function getQuestionId(question: string) {
+  let sanitizedQuestion = question;
+  // Remove trailing punctuation
+  sanitizedQuestion = sanitizedQuestion.replace(/[^\w\s]/g, "");
+  // Remove apostrophes
+  sanitizedQuestion = sanitizedQuestion.replace(/'/g, "");
+
+  return encodeURIComponent(sanitizedQuestion.toLowerCase().replace(/ /g, "-"));
+}
```
mbbischoff, line 39: Maybe we should consider using something we don’t need to transform for the IDs or even the index of question? Worried about the case where someone jams something else in here that isn’t URL safe.

mbbischoff:
Looks great. Ideas for design improvements that we might not care about now but might be ncie later:

* Animate the change to the next / previous question
* Indicate completed questions in the hamburger menu so you know what you’ve filled out and what you have left

src/components/forms/FormNavigation/FormNavigation.test.tsx
kydecker: Good call, I'll add those tests.

src/components/forms/FormSection/FormSection.tsx
kydecker: Yeah, I thought about that, too.

Right now all forms are coded manually and checked into the codebase (so there's no worry about unapproved user content for questions).

It might be good to eventually create a content management layer for forms that generates unique IDs, but that gets into "form builder CMS" territory that I think is worth avoiding right now due to complexity.

We **could** have an "id" prop required for every `FormSection` component, which would circumvent having to generate one. But it could get out of sync with the actual question, and would require some sort of URL validation anyways.

src/components/forms/FormNavigation/FormNavigation.test.tsx
kydecker: Ugh, I spent a few hours tonight trying without much luck.

In order to test link functionality, we need to mock TanStack Router (or parts of it). Documentation on how to do this is [spotty](https://github.com/TanStack/router/discussions/655).

There are some [examples from the TanStack repo itself](https://github.com/TanStack/router/blob/main/packages/react-router/tests/routeContext.test.tsx) that are worth exploring. I may try to do this with a little more rest in me first.

src/components/forms/FormSection/FormSection.tsx
mbbischoff: Totally fine to delay this to down the road. Just wanted to raise it.

src/components/forms/FormNavigation/FormNavigation.test.tsx
mbbischoff: Ugh, that’s so annoying. If you wanna split it out into a sepaprate ticket that’s fine

kydecker: Added a ticket: #405

mbbischoff (approved):

PR #403, opened 2025-02-18
feat: Create `HiddenText` component

### LINKED ISSUE #156: Create `HiddenText` component
Hide deadnames behind a spoiler-type mask after initial entry
Click-to-reveal

avevotsira:
## What changed?
I add a new component to hide and unhide text. 
 
You can use it with the following syntax:

```tsx
<HiddenText>John Doe</HiddenText>
```

## How was this tested?
I tested it manually and also wrote a simple test within: `src/components/common/HiddenText/HiddenText.test.tsx`


src/components/common/RedactedName/RedactedName.tsx
```diff
@@ -0,0 +1,48 @@
...
+      <TooltipTrigger>
+        <Button
+          variant="ghost"
+          size="small"
+          icon={isRevealed ? EyeOff : Eye}
+          onPress={() => setIsRevealed(!isRevealed)}
+          aria-label={isRevealed ? "Hide Deadname" : "Show Deadname"}
+          className="p-1.5"
+        />
+        <Tooltip>{isRevealed ? "Hide Deadname" : "Show Deadname"}</Tooltip>
```
kydecker:
```suggestion
          aria-label={isRevealed ? "Hide name" : "Show name"}
          className="p-1.5"
        />
        <Tooltip>{isRevealed ? "Hide name" : "Show name"}</Tooltip>
```
I'd prefer to use something more neutral than "deadname" for the label here.

The word "deadname" comes from the fact that, often times, when trans people would die, family would misgender and use their incorrect name during their funeral. 

Our trans people are living, so maybe just "Hide name" and "Show name"?

src/components/common/RedactedName/RedactedName.tsx
```diff
@@ -0,0 +1,48 @@
...
+      true: "bg-graya-2 dark:bg-graydarka-2",
+      false: "bg-graya-3 dark:bg-graydarka-3 text-transparent select-none",
+    },
+  },
+});
+
+export function RedactedName({
+  children,
+  className,
+}: RedactedNameProps): JSX.Element | null {
```
kydecker:
```suggestion
}: RedactedNameProps) {
```
nit: TypeScript should be able to infer this!

src/components/common/RedactedName/RedactedName.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const [isRevealed, setIsRevealed] = useState(false);
+
+  if (!children) return null;
+
+  return (
+    <span className={redactedStyles({ isRevealed, className })}>
+      <span className="px-1.5 py-0.5" aria-hidden={!isRevealed}>
```
kydecker:
```suggestion
      <span aria-hidden={!isRevealed}>
```
I'd like to make this component something that can be used inline with text, and adopt the same styles as the surrounding text!

So let's leave out any extra padding.

src/components/common/RedactedName/RedactedName.test.tsx
```diff
@@ -0,0 +1,40 @@
+import { render, screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it } from "vitest";
+import { RedactedName } from "./RedactedName";
+
+describe("RedactedName", () => {
```
kydecker: Nice! Thanks for writing tests!

kydecker (commented):
Hey @avevotsira! Thanks so much for your contribution! Really appreciate you diving in here and taking initiative on design and implementation.

I'd like to propose a few changes:

1. Ideally, we may use this to isolate a single name inside of a paragraph. With that in mind, I'd love to keep styling as minimal as possible here—no extra padding, backgrounds, or border radius. Just text! 
2. Maybe we can give this a similar treatment to spoiler text in other platforms like Discord—a black bar that, when clicked, reveals the text.

https://github.com/user-attachments/assets/e34a9f5a-2c3e-4ace-a3ae-01aeeeaa41fe

3. Can you create a Storybook story to demonstrate the functionality of this component? (This can help with development and testing, too!) You can do this by running `pnpm storybook`, opening the development server, and then clicking the "+" button to add a new story after selecting the `RedactedName.tsx` file.
4. To keep this generic, maybe we want to rename the component `RedactedText`? That way we can use it for things other than names, if necessary.

src/components/common/RedactedName/RedactedName.tsx
```diff
@@ -0,0 +1,48 @@
...
+export function RedactedName({
+  children,
+  className,
+}: RedactedNameProps): JSX.Element | null {
+  const [isRevealed, setIsRevealed] = useState(false);
+
+  if (!children) return null;
+
+  return (
+    <span className={redactedStyles({ isRevealed, className })}>
```
kydecker: It might be good to use the `<mark>` tag for this: https://www.w3schools.com/tags/tag_mark.asp

kydecker (commented):
Hey Ave!

I like the design of where this is headed. Thanks for your revisions.

I think we may be able to use a more accessible pattern that leverages the native HTML `checkbox` input. I put together a basic proof of concept in CodePen here: https://codepen.io/evadecker/pen/WbNGwrY

This should accomplish a few things:

1. By using `input type="checkbox"` instead of `button` we leverage more native HTML features, which means we can remove the custom keyboard listeners, and we get show/hide functionality "for free"
2. The `label` element will wrap to multiple lines automatically, which will fix this bug where the button is clickable outside of the spoiler area for multi-line spoilers

https://github.com/user-attachments/assets/08316130-0c65-490e-903a-3074edcf22cb

3. We'll need to do a little bit of extra work to show and hide the content from screen readers using `aria-label`. We want screen reader users to have a placeholder prompt like "click to reveal spoiler" in place of the actual text where possible.

I realize this is asking a lot so let me know if you need me to clarify anything, or want assistance with any part of this!

kydecker:
Hey @avevotsira! Thanks so much for all your work on this! It's looking great.

I just pushed a few tweaks to this branch which should get us over the finish line:

I appreciate you using aria live regions and going deep on accessibility! But when I tested it using VoiceOver, the messages were a little verbose, repeating things in the middle of a sentence. I tried to simplify the implementation and makes sure that the screen reader messages aren't repeated unnecessarily:

https://github.com/user-attachments/assets/0801cf0c-7aaa-4247-a06f-85de51825052

I also tweaked our handling of `onFocus` so that the focus rings will only display when users are navigating with the keyboard, and not on mouse click.

Finally, I thought it was better to rename this from `RedactedText` to `HiddenText` for simplicity.

Let me know if you have any thoughts or questions. Thanks again for your contributions on this!

kydecker (approved):

PR #411, opened 2025-02-25
chore: Make `quests.updatedAt` required

### LINKED ISSUE #393: Make Quests `updatedAt` required and migrate.

mbbischoff:
## What changed?
Fixes #393.

## Why?
We want all quests to have an updatedAt date, including when they’re created.

## How was this change made?
* I made updatedAt required and then updated the code to always pass it and test code to to the same.

## How was this tested?
1. Made a quest with an `undefine` updatedAt
2. Ran the migration via the dashboard
```
{
 Name: "migrations:setQuestsUpdatedTimestamp",
 Status:
"Migration was started and finished in one batch.",
 lastFinished: "2025-02-25T17:15:12.930Z",
 lastStarted: "2025-02-25T17:15:12.930Z",
 processed: 1,
 toStartOver: '{"cursor":null}',
}
```
3. Switched to this branch
4.  `pnpm test` & `pnpm dev`
5. Smoke tested the web app and admin side

## Anything else?
After this PR is approved, I can run the migration on the production database so that it’ll be ready for this deploy.


convex/migrations.test.ts
```diff
@@ -26,31 +26,6 @@ describe("migrations", () => {
...
-          ctx,
-          internal.migrations.setQuestsUpdatedTimestamp,
-        );
-
-        const updatedQuest = await ctx.db.get(questId);
-        expect(updatedQuest?.updatedAt).toBeDefined();
-        expect(typeof updatedQuest?.updatedAt).toBe("number");
-      });
-    });
-
```
mbbischoff, line 53: Removes this test because it no longer compiles once `updatedAt` is required.

src/components/common/FormattedDate/FormattedDate.test.tsx
```diff
@@ -1,7 +1,6 @@
 import { render, screen } from "@testing-library/react";
 import { afterEach, describe, expect, it } from "vitest";
 import { FormattedDate } from "./FormattedDate";
-
```
kydecker: Nit: keep this newline

src/components/common/index.ts
```diff
@@ -46,3 +46,5 @@ export * from "./ToggleButton/ToggleButton";
 export * from "./ToggleButtonGroup/ToggleButtonGroup";
 export * from "./Toolbar/Toolbar";
 export * from "./Tooltip/Tooltip";
+export * from "./FormattedDate/FormattedDate";
+export * from "./TimeAgo/TimeAgo";
```
kydecker: Nit: put these in alphabetic order

(I often `cmd` + `a` select all, then `cmd` + `shift` + `p` → "sort lines ascending" in VS Code)

kydecker (approved):
Seems straightforward enough!

PR #424, opened 2025-03-11
feat: Quests Completed Date

### LINKED ISSUE #186: Add "completed at" display for completed quests
Show start date and completion date, and difference between dates

mbbischoff:
## What changed?

Fixes #186.

## Why?
It’s nice to know when you finished something.

## How was this change made?
1. Added a startedAt property for quests and begins tracking it should we ever want to show that.
2. Added Footer UI for when a Quest was completed.

## How was this tested?

- [x] Started and finished Quests.
- [x] Ran tests

## Anything else?
I’ll also file a follow up for a nicer stamp-y design for this.


convex/userQuests.ts
```diff
@@ -148,26 +148,39 @@ export const setStatus = userMutation({
...
+      startedAt?: number | undefined;
+      completedAt?: number | undefined;
+    } = {
+      status: args.status as Status,
+    };
+
+    // Handle all status transitions
+    switch (args.status) {
+      case "notStarted":
+        update.startedAt = undefined;
```
kydecker: Do we want to unset `startedAt` here?

E.g. maybe a quest is "In Progress" and a user misclicks to "Not Started" instead of "Done". Now there's no way to set the start time back to the initial value.

Maybe we just count "startedAt" as the first time the quest is toggled away from "notStarted"?

src/components/common/Label/Label.tsx
```diff
@@ -0,0 +1,15 @@
+import type { LucideIcon } from "lucide-react";
+
+export interface LabelProps {
+  icon: LucideIcon;
+  children: React.ReactNode;
+}
+
+export function Label({ icon: Icon, children }: LabelProps) {
```
kydecker: `Label` to me implies a `label` element for forms — maybe there's a better name for this?

Alternatively we could just move this within `QuestPageFooter` as an internal component since we don't have anywhere else to use it right now.

We could also collapse `QuestCompletedDate` and `QuestUpdated` into `QuestPageFooter`, since they're all pretty small.

kydecker:
Will need a changeset here, too.

convex/userQuests.ts
mbbischoff: Yeah, I think it could make sense either way. But let’s go with your instinct here.

src/components/common/Label/Label.tsx
mbbischoff: Yeah, you’re right. Label is a silly name. That’s my [iOS dev life jumping out](https://developer.apple.com/documentation/swiftui/label).

kydecker: Interesting! So different from web.

convex/userQuests.ts
```diff
@@ -148,26 +148,38 @@ export const setStatus = userMutation({
...
     }
 
-    // If the status was already complete, unset completion time
-    if (userQuest.status === "complete") {
-      await ctx.db.patch(userQuest._id, {
-        status: args.status,
-        completedAt: undefined,
-      });
+    // Build the update object based on status transitions
+    const update: {
```
kydecker, line 156: Would be good to also update `userQuests.test.ts` with these new cases

src/components/common/IconText/index.ts
```diff
@@ -0,0 +1,2 @@
+export { IconText } from "./IconText";
+export type { IconTextProps } from "./IconText";
```
kydecker: Instead of having the `index.ts` file within the component directory can we add this to the barrel file at `common/index.ts`?

src/components/quests/QuestPageFooter/QuestPageFooter.tsx
```diff
@@ -1,14 +1,46 @@
+import { IconText } from "@/components/common/IconText";
+import { TimeAgo } from "@/components/common/TimeAgo/TimeAgo";
```
kydecker:
```suggestion
import { IconText, TimeAgo } from "@/components/common";
```
If we add to the barrel file

src/components/quests/QuestPageFooter/QuestPageFooter.tsx
```diff
@@ -1,14 +1,46 @@
...
   return (
-    <footer className="border-t border-gray-dim py-4">
+    <footer className="border-t border-gray-dim py-4 flex items-center">
+      <div className="flex-1">
+        {userQuest && <QuestCompletedDate userQuest={userQuest} />}
+      </div>
       <QuestUpdated quest={quest} />
     </footer>
   );
 }
```
kydecker: Nit: prefer the final component to come _after_ the definition any subcomponents

kydecker (approved):
Nice! LGTM! Yay tests!

Issue #410, opened 2025-02-25
Improve seed.ts to seed realistic data for dev databases


mbbischoff:
**Is your feature request related to a problem? Please describe.**

As a Namesake dev, it’d be faster to get up and running if when you started up the app, you already had a bunch of data to work with.

**Describe the solution you'd like**
I’d like for us to add more realistic data to the seed.ts file that’s automatically run in dev environments so that when I login as a developer for the first time, I have Quests, Steps, FAQs, etc to work with.

**Additional context**
[Discord](https://discord.com/channels/1250552190402035835/1259202186349514813/1343965867762778296)

PR #426, opened 2025-03-11
chore: Add script to delete all dev data

kydecker:
## What changed?
Add a new `pnpm dev:delete` script which will erase all Convex tables locally.

## Why?
When migrations occur, local dev data can get out of sync with production. Rather than attempt to set up migrations, it can be helpful to just delete and re-seed the dev database. See [YOLO: Get to an MVP fast](https://stack.convex.dev/yolo-fast-mvp#delete-dev-data-liberally-and-maintain-a-seed-script-to-re-initialize) and [Discord](https://discord.com/channels/1019350475847499849/1349141814703034508/1349145489970565162)

## How was this change made?
Added a new script to erase all Convex tables; added a new command that can be run from the CLI.

## How was this tested?
Ran it locally and it deleted everything!

## Anything else?
We should follow up with improvements to our `seed.ts` script that insert more data into the dev database. I updated #410 to move it onto the v1.0 board. It's not strictly necessary, definitely toward the bottom of the stack, but worth working on!

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "dev": "npm-run-all --parallel dev:frontend dev:backend",
     "dev:frontend": "vite",
     "dev:backend": "convex dev",
+    "dev:erase": "bash scripts/eraseDevData.sh",
     "build": "tsc -b && vite build",
     "lint": "biome lint",
     "lint:fix": "biome lint --write",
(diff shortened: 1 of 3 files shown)


mbbischoff:
great idea!

mbbischoff (approved):

PR #427, opened 2025-03-12
feat: Add `react-hook-form` and enable form submission

### LINKED ISSUE #325: Integrate `react-hook-form` and set up shared form management logic
Install [react-hook-form](https://react-hook-form.com/) to assist with validation and display.

Abstract and encapsulate component logic and shared props as-needed.

kydecker:
## What changed?
- Install [react-hook-form](https://react-hook-form.com/) to help manage form state
- Update all `/forms` input components to use RHF `Controller` to manage state
- Update MA court order form with additional questions and conditional show/hide logic for sub-questions
- Update form page design to use inline cards instead of full-screen "pages" for each section
- Add new `FormSubSection` component
- Change `/forms/court-order/ma` route to `/forms/ma-court-order`
- Resolves #325 

Data does not actually save to the backend right now, just outputs to `console.log()`. Will follow-up in #428 

https://github.com/user-attachments/assets/f8f385f6-507a-4511-afa5-5e43aa950549

## Why?
Progress toward saving user data to fill PDFs.

## How was this change made?
Added a new `UserFormDataFields` type to list all of the shared field names which can be used/reused across different forms. These types act as the keys in the `userFormData` table.

Installed react-hook-form, wrapped all form components in RHF `Controller`, changed the `name` prop to be required and take a `UserFormDataFields` type.

## How was this tested?
Updated existing test with a new `renderWithFormProvider` function. Made sure everything passed.

- [x] Update and test remaining `/forms` components
- [x] Add new tests to cover missed areas (RHF logic?)
- [x] Improve code coverage of new areas
- [x] Abstract some shared components? / code cleanup

## Anything else?
- Fixed alignment of the icon with `Banner` for `large` variants
- Support passing an `errorMessage` to an individual `Checkbox`
- Add `forwardRef` for most input components in order to enable RHF management and focus


mbbischoff:
Will try to review this by Friday if that’s alright.

kydecker:
> Will try to review this by Friday if that’s alright.

Yes, no rush! It's a large diff, but lots of repeated structures. Happy to answer any questions.

src/components/forms/AddressField/AddressField.test.tsx
```diff
@@ -57,23 +57,56 @@ describe("AddressField", () => {
...
 
   it("includes all autocomplete attributes", () => {
-    render(<AddressField />);
+    renderWithFormProvider(<AddressField type="residence" />);
 
     const streetAddressInput = screen.getByLabelText("Street address");
     const cityInput = screen.getByLabelText("City");
-    // const stateSelect = screen.getByLabelText("Statel");
+    const stateSelect = screen.getByRole("combobox", {
+      hidden: true,
```
mbbischoff, line 65: Hiding this since we already know their state for now?

src/components/forms/AddressField/AddressField.tsx
```diff
@@ -1,75 +1,133 @@
...
+    mailing: {
+      street: "mailingStreetAddress",
+      city: "mailingCity",
+      state: "mailingState",
+      zip: "mailingZipCode",
     },
-  });
+  };
+
+  const maskitoOptions: MaskitoOptions = {
```
mbbischoff, line 41: Maybe worth commenting what this is doing.

src/components/forms/FormNavigation/FormNavigation.tsx
```diff
@@ -90,7 +107,7 @@ export function FormNavigation() {
         >
           <ArrowUp className="size-5" />
         </Link>
-        <Tooltip>Previous question</Tooltip>
+        <Tooltip placement="left">Previous question</Tooltip>
```
mbbischoff, line 110: Should we use `start` rather than left just to set us up for RTL way down the road or does it not matter since a) that’s likely a long time from now and b) we could always grep for "left" / "right" anyway?

src/components/forms/FormSection/FormSection.tsx
```diff
@@ -44,28 +44,71 @@ export interface FormSectionProps extends FormHeaderProps {
...
       data-form-section
       data-testid="form-section"
       className={twMerge(
-        "flex flex-col gap-8 p-8 justify-center h-screen snap-center",
+        "flex flex-col gap-8 p-8 pb-9 justify-center outline outline-1 outline-graya-3 dark:outline-graydarka-3 shadow-sm dark:shadow-md rounded-2xl bg-white dark:bg-graydark-2",
         className,
       )}
+      // TODO: React-hook-form is supposed to prevent disabled fields from being submitted
+      // but it's not working for some reason
+      disabled={!isVisible}
```
mbbischoff: Should we ticket this?

src/components/forms/FormSection/FormSection.tsx
```diff
@@ -44,28 +44,71 @@ export interface FormSectionProps extends FormHeaderProps {
...
+}
+
+interface FormSubSectionProps {
+  title?: string;
+  children?: React.ReactNode;
+  className?: string;
+  isVisible?: boolean;
+}
+
+export function FormSubSection({
```
mbbischoff: Super minor, but I’d capitalize this as `FormSubsection`

src/components/forms/NameField/NameField.tsx
```diff
@@ -1,30 +1,51 @@
...
   return (
     <div className="@container flex flex-col gap-4">
       <div className="grid grid-cols-1 @lg:grid-cols-3 gap-4">
-        <TextField
-          label="First name"
-          name="firstName"
+        <ShortTextField
+          label="First or given name"
+          name={names[type].first}
           autoComplete="given-name"
```
mbbischoff, line 38: interesting to think about the autocomplete potential for name fields in the app since some browsers might let users only autocomplte one name and others may support picking from a list. I think you made the right call turning it on everywhere but I guess we might see that in feedback at some point.

src/components/forms/PhoneField/PhoneField.tsx
```diff
@@ -1,50 +1,66 @@
...
+  name,
+  defaultValue,
+  ...props
+}: PhoneFieldProps) {
+  const { control } = useFormContext();
+
+  const maskitoOptions: MaskitoOptions = {
+    mask: [
+      "+",
+      "1",
```
mbbischoff, line 22: Assuming this is the US country code, is it at all possible someone would fill out the form with a non-US number and that would be valid or do state courts only accept US numbers?

src/components/forms/AddressField/AddressField.test.tsx
kydecker: Ah, the native `select` input for React-Aria's `Select` component is hidden by default, so we have to target it using `hidden: true`.

But to your point, yes — we shouldn't need to ask for state here.

src/components/forms/FormNavigation/FormNavigation.tsx
kydecker: `placement` is a [standard prop from React Aria](https://react-spectrum.adobe.com/react-aria/Tooltip.html#tooltip-2) and they only use `"left | right | top | bottom"` so I think it's simpler to leave as-is.

src/components/forms/FormSection/FormSection.tsx
kydecker: Updated!

src/components/forms/NameField/NameField.tsx
kydecker: Yeah, good callout

src/components/forms/PhoneField/PhoneField.tsx
kydecker: That's a great question, I'm actually not sure. I assumed since Namesake only serves US residents that everyone will have a US number, but maybe that's an incorrect assumption

src/components/forms/FormSection/FormSection.tsx
kydecker: Planning to fix this when addressing the actual form submission in #428. Updated the comment to reference that ticket.

src/components/forms/AddressField/AddressField.tsx
kydecker: Added a comment!

src/components/forms/PhoneField/PhoneField.tsx
mbbischoff: We can roll with it for now and wait to get the feedback if you’d like?

kydecker: Yeah that works for me, my gut says it will be very rare that someone does not have a +1 phone number.

mbbischoff (approved):

Issue #161, opened 2024-10-28
Possible to make multiple new quest submissions on a slow connection


evadecker:
Throttle new quest additions / disable button on click to prevent duplicate submissions

mbbischoff:
Before writing code here, make sure to check that it wasn’t already resolved

PR #432, opened 2025-03-14
feat: Update design of data settings page, rename to "Form Responses"

### LINKED ISSUE #326: Allow viewing and deleting form responses from settings
Values saved to `userFormData` should be shown to that user in a table within the /data page of Settings.

Allow users to delete each row individually, or delete all data at once.

kydecker:
## What changed?
- Rename `/settings/data` to `/settings/responses` and "Data" to "Form Responses"
- Rename `userFormData` table to `userFormResponses`
- Display an "Encrypted" badge in the page header
- Display form responses using a `ListBox`
- Allow deleting responses individually or in bulk
- Display an empty state when no responses exist
- Only allow admins to view response entry form (mostly used for testing)
- Add a new `useEncryptionKey()` hook to simplify fetching a user's encryption key for data
- Resolves #326 

CleanShot 2025-03-14 at 15 20 58@2x

https://github.com/user-attachments/assets/0d112948-2a6e-481b-af8c-37826d4e1da5

## Why?
The existing design and implementation was a placeholder only, used to demonstrate how to encrypt and decrypt user data. This improves the design 

## How was this tested?
Wrote unit tests for `FormResponsesList` and `DeleteFormResponseModal`. Manually tested adding and deleting data from the table.

## Anything else?
#427 introduces a global `const` for tracking all of the potential keys for user form responses. Once that has merged, we could change the text input here to a `Select` dropdown for choosing the field name. We should also be able to display a human-readable label instead of the field ID itself. (For example, "First Name" instead of "firstName").

Although the encrypted values are all stored as strings, we may also be able to store a data type and modify the UI based on the type of data stored (e.g. boolean, string, array).


mbbischoff:
Design preference here, I’d tighten up the vertical spacing between "No data" and the "Form resposnes will appear here" text a little. Right now it looks like there’s equal spacing between the icon and both runs of text and I think it’d look more harmonious with some grouping.

convex/userFormResponses.ts
```diff
@@ -44,10 +44,35 @@ export const set = userMutation({
...
-    await ctx.db.insert("userFormData", {
+    await ctx.db.insert("userFormResponses", {
       userId: ctx.userId,
       field: args.field,
       value: args.value,
     });
   },
 });
+
+export const deleteAll = userMutation({
```
mbbischoff: This name at first made me feel like it’ll delete all `userFormResponses`, not just the current user’s.

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+  const deleteLabel = getDeleteLabel({
+    selectedRows,
+    totalCount: rows?.length ?? 0,
+  });
+
+  const handleDelete = async () => {
+    try {
+      setIsDeleting(true);
+
+      if (selectedRows === "all") {
```
mbbischoff: Should we maybe have a constant for this "all". Seeing it as a string literal multiple places makes me a little nervous.

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+        await deleteByIds({
+          userFormResponseIds: Array.from(
+            selectedRows,
+          ) as Id<"userFormResponses">[],
+        });
+      }
+
+      onOpenChange(false);
+      onDelete();
+      toast.success("Data deleted");
```
mbbischoff:
```suggestion
      toast.success("Response deleted");
```

Maybe for consistency?

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+    }
+  };
+
+  return (
+    <Modal isOpen={isOpen} onOpenChange={onOpenChange}>
+      <ModalHeader title={`${deleteLabel}?`} />
+      {error && <Banner variant="danger">{error}</Banner>}
+      {selectedRows !== "all" && (
+        <>
+          <p>The following data will be deleted:</p>
```
mbbischoff:
```suggestion
          <p>The following responses will be deleted:</p>
```

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+          <ul className="list-disc list-inside text-sm leading-6 text-gray-dim">
+            {Array.from(selectedRows).map((row) => (
+              <li key={row}>{rows?.find((r) => r.id === row)?.field}</li>
+            ))}
+          </ul>
+        </>
+      )}
+      <Banner variant="warning">This action cannot be undone.</Banner>
```
mbbischoff:
```suggestion
      <Banner variant="warning">Deleting responses cannot be undone.</Banner>
```

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
```diff
@@ -0,0 +1,101 @@
...
+        >
+          Cancel
+        </Button>
+        <Button
+          variant="destructive"
+          onPress={handleDelete}
+          isDisabled={isDeleting}
+        >
+          {isDeleting ? "Deleting…" : deleteLabel}
+        </Button>
```
mbbischoff: Do we want to have a progress state for our buttons or do we want spinners to be in the top right of our modals? Thinkning about this for https://github.com/namesakefyi/namesake/issues/161 also.

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
kydecker: Yeah, `"all"` is part of how React Aria handles [select all](https://react-spectrum.adobe.com/react-aria/selection.html#select-all) in order to allow interacting with elements in a list that aren't loaded locally.

FWIW — TypeScript will fail to compile if you put anything but `"all"` here, or if you try to access the selected elements without checking for `"all"`.

Do you still think it's helpful to have a constant? Just `export const ALL = "all"`?

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
kydecker: Mm, it might be good to have a built-in `isLoading` state for buttons?

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
kydecker: True. Changed the `getDeleteLabel` function to `getResponseCountLabel` so it can be reused to pluralize the confirmation here.

kydecker:
> Design preference here, I’d tighten up the vertical spacing between "No data" and the "Form resposnes will appear here" text a little. Right now it looks like there’s equal spacing between the icon and both runs of text and I think it’d look more harmonious with some grouping.

Tightened!

CleanShot 2025-03-15 at 22 06 55@2x

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
kydecker: Made `all` a const: https://github.com/namesakefyi/namesake/pull/432/commits/2ef941051d03581a3d6de66ab7f68701ccf025a5

convex/userFormResponses.ts
kydecker: Renamed to `deleteAllForCurrentUser`: https://github.com/namesakefyi/namesake/pull/432/commits/1cdd54968d6033202e0d6e8c6aed10c9d2ccbb4a

src/components/settings/DeleteFormResponseModal/DeleteFormResponseModal.tsx
mbbischoff: Cool, I’ll tackle that in another PR

mbbischoff (approved):

Issue #439, opened 2025-03-17
Use structured fields for form response `value` entry


evadecker:
**Is your feature request related to a problem? Please describe.**
Currently, values in `/settings/responses` accept plain text input, but some fields store a boolean, an array, or other structured data.

**Describe the solution you'd like**
It would be nice to display UI which is associated with the stored data type. (E.g. a checkbox or a switch for `boolean`, a checkbox group or dropdown for `array`, etc.)

**Describe alternatives you've considered**
N/A

PR #440, opened 2025-03-17
feat: Use `Select` for Form Response values

kydecker:
## What changed?
Followup to #432. Use the constants for field names to populate a human-readable `Select` dropdown instead of displaying a plain text input.

https://github.com/user-attachments/assets/d6a2ab9c-099e-4f31-a7a8-d92269b530e7

## Why?
More structured UI with fewer opportunities for user error, typos, etc.

## How was this change made?
Replaced the existing `TextField` with a `Select` and updated relevant handlers and TypeScript types. Added a new function for `getReadableFieldLabel()`.

## How was this tested?
Wrote tests for new function, verified that values display properly in the settings page.

## Anything else?
Would be nice to set up response fields with more structured types, e.g. to declare that `isUnhoused` is a `boolean`. This way, instead of providing a plain `TextField` for each form response, we can provide a more tailored input. This is not necessary for v1 but I'll file a ticket in the backlog: #439


src/components/settings/FormResponsesList/FormResponsesList.tsx
```diff
@@ -122,13 +143,23 @@ function AddFormResponse() {
       onSubmit={handleSubmit}
       className="border border-gray-dim rounded-lg p-4 flex-row gap-2 items-end"
     >
-      <TextField
+      <Select
         label="Field"
```
mbbischoff:
```suggestion
        label="Question"
```

src/components/settings/FormResponsesList/FormResponsesList.tsx
```diff
@@ -122,13 +143,23 @@ function AddFormResponse() {
...
+        onSelectionChange={(key) => setField(key as UserFormDataField)}
+      >
+        {(item) => (
+          <SelectItem key={item.value} id={item.value}>
+            {item.label}
+          </SelectItem>
+        )}
+      </Select>
       <TextField
         label="Value"
```
mbbischoff:
```suggestion
        label="Response"
```

src/components/settings/FormResponsesList/FormResponsesList.tsx
```diff
@@ -139,7 +170,7 @@ function AddFormResponse() {
       <Button
         type="submit"
         variant="secondary"
-        isDisabled={field === "" || value === "" || isSaving}
+        isDisabled={!field || value === "" || isSaving}
       >
         {isSaving ? "Saving..." : "Add response"}
```
mbbischoff:
```suggestion
        {isSaving ? "Saving..." : "Save"}
```

mbbischoff (approved):
Code looks great, functionality looks great, some small copy suggestions to make things a little less programmer-y that you can feel free to take or leave.

kydecker:
> Code looks great, functionality looks great, some small copy suggestions to make things a little less programmer-y that you can feel free to take or leave.

Great callout 🙏 Changed!

PR #443, opened 2025-03-19
fix: Unknown `isReadOnly` prop and other React console errors

### LINKED ISSUE #430: Console errors for `isReadOnly`, `isRequired`, and `defaultChildren` props
**Describe the bug**
React is displaying console errors for certain props. These errors are also visible when running `pnpm test`.

All the props named in the errors are from React Aria, so we probably are passing down props in certain places we shouldn't be.

**To Reproduce**
1. Go to a page with inputs like `/forms/ma-court-order`
2. Open the terminal
3. Observe errors

**Expected behavior**
No errors shown.

**Screenshots**
Image

 - OS: macOS
 - Browser Chrome

kydecker:
## What changed?
Fixes #430.

## How was this change made?
Destructure all `renderProps` that get passed into `FieldGroup` to only pass down supported props.

## How was this tested?
Opened all common components in Storybook to verify no console errors display. Ran `pnpm test` and verified no errors or warnings emit.

### DIFF
--- a/.changeset/green-ends-invent.md
+++ b/.changeset/green-ends-invent.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Fix React console errors about invalid props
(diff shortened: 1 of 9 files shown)


mbbischoff (approved):
Thank you for fixing this! It was driving me a little batty 🦇

PR #444, opened 2025-03-19
feat: Save and encrypt form submissions

kydecker:
## What changed?
Save and encrypt all user data at the end of a form. Resolves #428.

https://github.com/user-attachments/assets/7d709242-49dc-4e17-b25b-18d272420c23

## Why?
Saved data will be stored in Convex and used to fill out PDFs and other forms.

## How was this change made?
- Created a new `useFormEncryptAndSubmit` hook to handle form submission in a reusable way
- Implemented the hook on our first form, `ma-court-order`
- Verified that data is properly encrypted, submitted, and stored
- Verified that data displays on the form responses list in settings

## How was this tested?
Wrote new unit tests for the `useFormEncryptAndSubmit` hook.

## Anything else?
- Add a new capitalization helper and fix an issue where the "all responses deleted" was not properly capitalized.
- Extend some test mocks

Issue #116, opened 2024-09-30
Improve error boundaries and error handling


evadecker:
- [x] Use `ErrorBoundary`, `Suspense`, etc. to prevent the app from breaking when something doesn't render correctly.
- [x] Test by forcing different parts of the app into an error state and ensuring that most of the app still renders.
- [x] Exceptions should be reported to PostHog

PR #450, opened 2025-03-26
feat: Prefill form fields from existing responses

### LINKED ISSUE #327: Pre-fill existing data to new forms when available
If a user has previously entered data for e.g. new name, address, email, we should use that data to pre-fill any new forms they encounter.

There should be a visual treatment to indicate that the form has been pre-filled with existing data (e.g. yellow background).

kydecker:
## What changed?
Pre-fill form fields from existing user responses. Resolves #327.

https://github.com/user-attachments/assets/4accd713-2564-43d1-8c58-3deca83c7093

## Why?
Prevent users from tedious data re-entry.

## How was this change made?
- Updated the `USER_FORM_DATA_FIELDS` const to store a `type` field for inferring types
- Add new `getByFields` query on the `UserFormResponses` table
- Create new `useDecrypt` hook to abstract away decryption logic from the `FormResponsesList` component
- Create a new `useForm` hook which handles a) decrypting existing fields, b) passing them to React Hook Form as default values, and c) encrypting and saving submissions to the database. The hook extends all properties available from RHF's own `useForm` and should be seen as our own custom replacement. This replaces and extends the existing `useFormEncryptAndSubmit` hook, which has been deleted.

## How was this tested?
Wrote tests for `useDecrypt` and `useForm` hooks.

## Anything else?
- In order to store various datatypes reliably as encrypted strings, the core encrypt and decrypt functions were updated to use `JSON.stringify` (before encrypting) and `JSON.parse` (after decrypting).
- Updated all `form.watch()` calls to explicitly check for `=== true`. It seems like unmounting the form fields introduced in #444 changed the way `.watch()` works.


mbbischoff:
Testing on this branch nothing happens when I click Submit on http://localhost:5173/forms/ma-court-order. Is there something else I need to setup locally first?

src/utils/encryption.ts
```diff
@@ -46,7 +44,8 @@ export async function decryptData(
     dek,
     data,
   );
-  return new TextDecoder().decode(decryptedData);
+  const decryptedString = new TextDecoder().decode(decryptedData);
+  return JSON.parse(decryptedString);
```
mbbischoff, line 48: Do we need to worry about the possibility of a SyntaxError being thrown here if we accidentally write bad JSON at some point?

src/utils/encryption.ts
```diff
@@ -163,3 +162,74 @@ async function retrieveDEK(): Promise<string | null> {
...
+};
+
+type DecryptResult = SingleDecryptResult | MultipleDecryptResult;
+
+/**
+ * Decrypts a single or multiple encrypted values.
+ * @param encryptedValue - The encrypted value to decrypt.
+ * @returns The decrypted value or values.
+ */
+export function useDecrypt(
```
mbbischoff, line 186: Really nice!

kydecker:
@mbbischoff There shouldn't be any extra setup needed! Try installing dependencies and checking the console.

Right now submitting the form should pop up a toast but not navigate to a new page.

src/utils/encryption.ts
kydecker: Do you mean if JSON.stringify() fails? Not sure what happens then.

We aren't writing JSON manually, this adds stringification of data before encryption and then reverses it during decryption.

submit(data) → JSON.stringify(data) → encrypt(data)

decrypt(encryptedData) → JSON.parse(data) → data

Using JSON.stringify() and parse() should be more robust than the previous approach of calling toString() on non-string values.

mbbischoff:
Hitting this error when trying to submit the form:

```
[Error] No encryption key available
	(anonymous function) (useForm.ts:42)
	(anonymous function) (useForm.ts:40)
	(anonymous function) (react-hook-form.js:1683)
```

src/utils/encryption.ts
mbbischoff: Makes sense. Maybe when we tackle https://github.com/namesakefyi/namesake/issues/116, we can get a little more defensive about writing / parse errors to account for the possibility of someone shipping bad code, but for now I think this is fine.

convex/constants.ts
```diff
@@ -341,39 +341,145 @@ export const COMMON_PRONOUNS = [
...
 export type UserFormDataField = keyof typeof USER_FORM_DATA_FIELDS;
+
+export type FieldType<K extends UserFormDataField> =
+  (typeof USER_FORM_DATA_FIELDS)[K]["type"] extends "boolean"
+    ? boolean
+    : (typeof USER_FORM_DATA_FIELDS)[K]["type"] extends "string[]"
+      ? string[]
+      : (typeof USER_FORM_DATA_FIELDS)[K]["type"] extends "string"
+        ? string
+        : any;
```
mbbischoff:
```suggestion
export type FieldType<K extends UserFormDataField> = {
  boolean: boolean;
  "string[]": string[];
  string: string;
}[(typeof USER_FORM_DATA_FIELDS)[K]["type"]];
```

Maybe reads a little nicer and lets us remove the `any`

src/routes/_authenticated/forms/ma-court-order.tsx
mbbischoff, line 1: On line 111 should we also do an === check for `isCurrentlyUnhoused`

kydecker:
@mbbischoff Pull and try again. Should be fixed.

Updated https://github.com/namesakefyi/namesake/pull/450/commits/7ed930d86f2a627fb0456ab8edf14c4bf143cbb6#diff-fa4ba159ff33a29146ff3f9078602be378024350e1cdce4dd1b7b837a2c1fbcbR76 to initialize a new encryption key if one is not already available.

src/utils/encryption.ts
```diff
@@ -163,3 +162,74 @@ async function retrieveDEK(): Promise<string | null> {
...
+  decryptedValue: string | undefined;
+  decryptedValues?: never;
+  error: boolean;
+};
+
+type MultipleDecryptResult = {
+  decryptedValue?: never;
+  decryptedValues: string[] | undefined;
+  error: boolean;
+};
```
mbbischoff, line 177: What if we did something like

```typescript
 interface DecryptionResult {
    error: boolean;
  }

  interface SingleDecryptResult extends DecryptionResult {
    type: 'single';
    decryptedValue: string | null;
  }

  interface MultipleDecryptResult extends DecryptionResult {
    type: 'multiple';
    decryptedValues: string[];
  }
  ```

Which would maybe let us remove the type assertions of `as MultipleDecryptResult` and `as SingleDecryptResult` below.

mbbischoff:
> @mbbischoff Pull and try again. Should be fixed.
> 
> Updated [7ed930d#diff-fa4ba159ff33a29146ff3f9078602be378024350e1cdce4dd1b7b837a2c1fbcbR76](https://github.com/namesakefyi/namesake/commit/7ed930d86f2a627fb0456ab8edf14c4bf143cbb6#diff-fa4ba159ff33a29146ff3f9078602be378024350e1cdce4dd1b7b837a2c1fbcbR76) to initialize a new encryption key if one is not already available.

Works great

convex/constants.ts
kydecker: Oh, this is *much* nicer.

src/utils/encryption.ts
kydecker: This would require typing the return value of `useDecrypt` as `SingleDecryptResult | MultipleDecryptResult`, which feels to me more confusing than a single `DecryptResult` type which can handle either multiple or singular values.

I tried removing the type assertions at the bottom with the existing types and it works; I'm good with that if you are!

mbbischoff: Okay that’s good with me

mbbischoff (approved):

PR #456, opened 2025-03-28
style: Fix `SkeletonCircle` rounded pseudo style

fartinmartin:
## What changed?
Skeleton animation is styled via a `:before` pseudo element instead of the root `<div>`, so the `rounded-full` class on the circle variant needs to target the pseudo element. Closes #445.


kydecker (approved):
Love an easy fix. Thank you, @fartinmartin!

PR #466, opened 2025-04-03
feat: Make app responsive

### LINKED ISSUE #182: Make app mobile-responsive
Sidebar should collapse into left pane and display a menu to expand/collapse. Tap targets should all be touch-friendly.

kydecker:
## What changed?
Make app mobile-friendly. Resolves #182

https://github.com/user-attachments/assets/a8a9f001-2b2b-4bf7-baec-a4c70ff99cbf

https://github.com/user-attachments/assets/1e626a04-785c-48a8-b0e2-0db0cc215262

Videos shown on iPhone SE.

Additionally:
- Relocate `/$questSlug` route to `/quests/$questSlug` to prevent polluting the top-level routes and have more control over empty state rendering
- Remove the card containers from form sections to give more space for text on mobile; subsections are now indented on the left to communicate the relation to the parent section
- Remove the user menu dropdown from the main app sidebar and replace with a link to `/settings` and a link to `/admin` (if the user is an admin)
- Add a "Sign out" link to the settings nav (necessary for mobile due to lack of user menu dropdown) and add a new "/signout" route

CleanShot 2025-04-03 at 15 49 21@2x

## Why?
Phones! People use em!

## How was this change made?
- Created a new `useIsMobile()` hook which can be used to test a constant breakpoint across the app
- Split apart `AppSidebar` component into `AppSidebarHeader`, `AppSidebar, and `AppSidebarFooter` in order to permit reusing sidebar logic when certain elements are not needed
- Added new `AppMobileNav` to display at the bottom of the viewport
- Added a `backLink` prop to `PageHeader` to support displaying a back arrow
- Modified routing logic to conditionally render the sidebar navigation as an index route when there isn't enough space for a sidebar
  - For example, on desktop, `/settings` will automatically route the user to `/settings/account`, showing the settings sidebar alongside the account settings; on mobile, `/settings` will render the settings sidebar nav as a full-page view of its own

## How was this tested?
Updated tests and wrote a few more, mostly manual clicking around the app locally and opening the mobile emulator.

## Anything else?
The quest page header needs some work; I'm going to follow up with a separate PR to improve that since this one has already gotten large. Filed #467

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,16 +1,21 @@
 <!doctype html>
 <html lang="en">
-  <head>
-    <meta charset="UTF-8" />
-    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
-    <link rel="icon" href="/favicon.ico" sizes="32x32" />
-    <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
-    <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
-    <link rel="manifest" crossorigin="use-credentials" href="/manifest.webmanifest" />
-    <title>Namesake</title>
-  </head>
-  <body>
-    <div id="root"></div>
-    <script type="module" src="/src/main.tsx"></script>
-  </body>
+
+<head>
+  <meta charset="UTF-8" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, viewport-fit=cover" />
+  <meta name="theme-color" content="#fcfcfd" media="(prefers-color-scheme: light)" />
+  <meta name="theme-color" content="#18191b" media="(prefers-color-scheme: dark)" />
+  <link rel="icon" href="/favicon.ico" sizes="32x32" />
+  <link rel="icon" type="image/svg+xml" href="/favicon.svg" />
+  <link rel="apple-touch-icon" href="/apple-touch-icon.png" />
+  <link rel="manifest" crossorigin="use-credentials" href="/manifest.webmanifest" />
+  <title>Namesake</title>
+</head>
+
+<body>
+  <div id="root"></div>
+  <script type="module" src="/src/main.tsx"></script>
+</body>
+
 </html>
--- a/public/manifest.webmanifest
+++ b/public/manifest.webmanifest
@@ -3,5 +3,6 @@
   "icons": [
     { "src": "/favicon-192.png", "type": "image/png", "sizes": "192x192" },
     { "src": "/favicon-512.png", "type": "image/png", "sizes": "512x512" }
-  ]
+  ],
+  "display": "standalone"
 }
--- a/src/components/app/AppMobileNav/AppMobileNav.tsx
+++ b/src/components/app/AppMobileNav/AppMobileNav.tsx
@@ -25,7 +25,14 @@ const AppMobileNavItem = ({
   ...props
 }: AppMobileNavItemProps) => {
   const matchRoute = useMatchRoute();
-  const isActive = Boolean(matchRoute({ ...props.href, fuzzy: true }));
+
+  const isActive =
+    label === "Home"
+      ? Boolean(
+          matchRoute({ to: "/", fuzzy: true }) ||
+            matchRoute({ to: "/quests/$questSlug", fuzzy: true }),
+        )
+      : Boolean(matchRoute({ ...props.href, fuzzy: true }));
 
   return (
     <Link {...props} className={navItemStyles({ isActive })}>
--- a/src/components/app/AppMobileNav/AppMobileNav.tsx
+++ b/src/components/app/AppMobileNav/AppMobileNav.tsx
@@ -1,6 +1,8 @@
 import { Link, type LinkProps } from "@/components/common";
+import { api } from "@convex/_generated/api";
 import { type ToOptions, useMatchRoute } from "@tanstack/react-router";
-import { Home, Settings } from "lucide-react";
+import { useQuery } from "convex/react";
+import { GlobeLock, Home, Settings } from "lucide-react";
 import type { LucideIcon } from "lucide-react";
 import { tv } from "tailwind-variants";
 
@@ -43,14 +45,24 @@ const AppMobileNavItem = ({
 };
 
 export const AppMobileNav = () => {
+  const user = useQuery(api.users.getCurrent);
+  const isAdmin = user?.role === "admin";
+
   return (
-    <div className="w-full shrink-0 bg-element border-t border-gray-a4 grid grid-cols-2 gap-1 p-2 items-center sticky bottom-0 mt-auto shadow-2xl">
+    <div className="w-full shrink-0 bg-element border-t border-gray-a4 flex *:flex-1 *:shrink-0 gap-1 p-2 items-center sticky bottom-0 mt-auto shadow-2xl">
       <AppMobileNavItem icon={Home} label="Home" href={{ to: "/" }} />
       <AppMobileNavItem
         icon={Settings}
         label="Settings"
         href={{ to: "/settings" }}
       />
+      {isAdmin && (
+        <AppMobileNavItem
+          icon={GlobeLock}
+          label="Admin"
+          href={{ to: "/admin" }}
+        />
+      )}
     </div>
   );
 };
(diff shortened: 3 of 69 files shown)


mbbischoff:
Taking a testing pass first. And then once we talk through / file any of these items, I’ll look at code.

## Design Notes

image

Maybe we can decrease the logo vertical padding here so that the username and password field both fully fit on screen with the keyboard?

* Perhaps we should app a web app manifest so that the Safari UI can be hidden when installed to the home screen?

Simulator Screenshot - iPhone 16 - 2025-04-06 at 21 45 35

* Can we make it so the tab bar sticks at the bottom and doesn’t get pulled along with the scroll?

Simulator Screenshot - iPhone 16 - 2025-04-06 at 21 46 11

* The back icon doesn’t make a ton of sense here because I tapped on the tab to get to Admin.


Simulator Screenshot - iPhone 16 - 2025-04-06 at 21 46 35

* New quest button is offscreen.

Simulator Screenshot - iPhone 16 - 2025-04-06 at 21 47 10

* Same for the early access generate code button.

Simulator Screenshot - iPhone 16 - 2025-04-06 at 21 47 52
 
* Editing a Quest expands the layout horizontally unexpectedly.

* Tapping on a quest gets you to the quest but without a back button to get to the sidebar.


image


* Sometimes tapping Admin takes you to a blank screen and you have to scroll up to see the content.

image

* Maybe these editable fields should go below on mobile?

kydecker:
> Maybe we can decrease the logo vertical padding here so that the username and password field both fully fit on screen with the keyboard?

Done ✅ 

Simulator Screenshot - iPhone 15 Pro - 2025-04-08 at 09 34 50

> Perhaps we should app a web app manifest so that the Safari UI can be hidden when installed to the home screen?

Done ✅ 

Simulator Screenshot - iPhone 15 Pro - 2025-04-08 at 11 44 13

> Can we make it so the tab bar sticks at the bottom and doesn’t get pulled along with the scroll?

Tab bar is now position Fixed ✅ 

https://github.com/user-attachments/assets/a2be872b-fd70-481e-ac15-8c1789023e2e

> The back icon doesn’t make a ton of sense here because I tapped on the tab to get to Admin.

Removed top-level back buttons ✅ 

> New quest button is offscreen.

> Same for the early access generate code button.

Might wait for a follow-up to fix these, mobile admin screens feel low priority.

> Editing a Quest expands the layout horizontally unexpectedly.

> Tapping on a quest gets you to the quest but without a back button to get to the sidebar.

These both need some more design thought. Will tackle in #467.

> Sometimes tapping Admin takes you to a blank screen and you have to scroll up to see the content.

A minimum page height was being set which didn't account for the tab bar, fixed ✅

> Maybe these editable fields should go below on mobile?

Toyed with this for a little bit and couldn't get the inline form to feel natural. Put it in a Popover and I think it works well. What do you think?

CleanShot 2025-04-08 at 11 08 06@2x

mbbischoff:
> Toyed with this for a little bit and couldn't get the inline form to feel natural. Put it in a Popover and I think it works well. What do you think?

Good idea. I like that!

index.html
```diff
@@ -1,16 +1,21 @@
...
+  <link rel="manifest" crossorigin="use-credentials" href="/manifest.webmanifest" />
+  <title>Namesake</title>
+</head>
+
+<body>
+  <div id="root"></div>
+  <script type="module" src="/src/main.tsx"></script>
+</body>
+
+</html>
```
mbbischoff, line 21:
```suggestion
</html>

```

public/manifest.webmanifest
```diff
@@ -3,5 +3,6 @@
   "icons": [
     { "src": "/favicon-192.png", "type": "image/png", "sizes": "192x192" },
     { "src": "/favicon-512.png", "type": "image/png", "sizes": "512x512" }
-  ]
+  ],
+  "display": "standalone"
```
mbbischoff, line 7: Nice

src/components/app/AppMobileNav/AppMobileNav.tsx
```diff
@@ -0,0 +1,78 @@
...
+
+const AppMobileNavItem = ({
+  icon: Icon,
+  label,
+  ...props
+}: AppMobileNavItemProps) => {
+  const matchRoute = useMatchRoute();
+
+  const isActive =
+    label === "Home"
```
mbbischoff, line 42: Should we define this as a constant just in case we change it later to someting like "Dashboard" or whatever so we don’t have to hunt it down in a few spots?

src/components/app/AppMobileNav/AppMobileNav.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <Link {...props} className={navItemStyles({ isActive })}>
+      <Icon className={iconStyles({ isActive })} />
+      {label}
+    </Link>
+  );
+};
+
+export const AppMobileNav = () => {
+  const user = useQuery(api.users.getCurrent);
+  const isAdmin = user?.role === "admin";
```
mbbischoff, line 59: Makes me think it might be nice to have a `user.isAdmin` that just checks role internally?

src/styles/index.css
```diff
@@ -7,13 +7,18 @@
...
-  @apply px-6 lg:px-8;
+  @apply px-4 md:px-6 lg:px-8;
 }
 
 .h-header {
   @apply h-16 lg:h-20;
 }
 
+.h-mobile-nav {
+  height: calc(56px + env(safe-area-inset-bottom));
```
mbbischoff, line 18: Where does the 56px come from?

src/routes/_authenticated/_home/quests.$questSlug.index.tsx
```diff
@@ -40,13 +37,20 @@ export const Route = createFileRoute("/_authenticated/_home/$questSlug/")({
...
 
   // TODO: Improve loading state to prevent flash of empty
   if (questData === undefined) return;
   if (questData.quest === null)
-    return <Empty title="Quest not found" icon={Milestone} />;
+    return (
+      <Empty
+        title="Quest not found"
+        icon={Milestone}
+        button={{ children: "Go back", onPress: () => router.history.go(-1) }}
```
mbbischoff: Confirming that `router.history.go(-1)` is safe even if there’s nothing in the history.

mbbischoff (approved):
Approved with minor comments.

src/styles/index.css
kydecker: I tried to match the typical height of an iOS tab bar! I remember reading it was `56px` but now I'm finding conflicting info about that.

mbbischoff: It’s 49 + safe area

mbbischoff: ignore my last comment, you’re right. The app I use for this is broken currently. 56 px + safe area is good.

index.html
kydecker: Trying to save this and the formatter keeps removing the blank EOL. Maybe because HTML is space-sensitive? Or maybe my editor settings need adjusting.

src/components/app/AppMobileNav/AppMobileNav.tsx
kydecker: I would love to rethink how we are fetching roles like this entirely. I don't love needing a query in the component just to get back a role. Maybe we can set up a provider for context like this. Some of it's already saved in TanStack router and accessible before load, but not sure how we pass it down to a component. Seems like a good spot for a new hook?

Gonna save this for later: https://github.com/namesakefyi/namesake/issues/476

src/routes/_authenticated/_home/quests.$questSlug.index.tsx
kydecker: Replaced with a home link since the back button can take you away from Namesake if this is the first page you've linked to.

index.html
mbbischoff: Let’s not worry about it then. It’s not a huge deal.

src/components/app/AppMobileNav/AppMobileNav.tsx
mbbischoff: Sounds good

src/components/app/AppMobileNav/AppMobileNav.tsx
kydecker: Attempted refactoring this out into a const for all the nav items, but doing that was a little awkward without easier access to `isAdmin`, so I'm gonna leave this as-is and put a note in the other ticket.

PR #469, opened 2025-04-07
feat: Support showing loading state in `Button`

### LINKED ISSUE #161: Possible to make multiple new quest submissions on a slow connection

mbbischoff:
## What changed?
 
* Adds a progress state to buttons and employs it where it makes sense.

https://github.com/user-attachments/assets/2b72fa00-3f14-4a2f-9425-acb37ccf056e

Fixes #161

## Why?
* Shows the user progress and makes sure the button is disabled when it should be.

## How was this change made?
* Added the new state showing a loader
* Added storyboook stories
* Added tests

## How was this tested?
Aritificially introduced delays to see the spinner status.

## Anything else?
I could use some help firguring out the best way to animate the button width changing because it feels a little jumpy right now. The `transition-all` didn’t quite work the way I hoped it would.


.changeset/smooth-students-knock.md
```diff
@@ -0,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Add loading indicators to buttons during form submission to improve user experience
```
kydecker:
```suggestion
Add loading indicators to buttons during form submission
```
Nit

src/components/common/Button/Button.tsx
```diff
@@ -71,35 +82,44 @@ export function Button({
...
+          {children}
+        </>
+      ) : (
+        <>
+          {Icon && (
+            <Icon {...sharedIconProps} className="shrink-0" {...iconProps} />
+          )}
+          {children}
+          {EndIcon && <EndIcon {...sharedIconProps} {...endIconProps} />}
+        </>
```
kydecker: This will cause the button to reflow/resize in some cases:

- original button has no icon
- original button has end icon

I wonder if instead, we:

1. Visually hide the existing contents of the button
2. Overlay a centered spinner

Radix uses this approach, see: https://www.radix-ui.com/themes/docs/components/button#loading

CleanShot 2025-04-08 at 15 28 46@2x

This will simplify some of the layout considerations for the button! Lmk what you think.

src/components/common/Button/Button.tsx
```diff
@@ -55,6 +61,11 @@ export const buttonStyles = tv({
       size: "medium",
       className: "w-10 h-10 p-2",
     },
+    {
+      isDisabled: true,
+      isSubmitting: false,
+      className: "opacity-40",
```
kydecker: What's the thought behind diverging from disabled state and submitting (disabled implied) here?

kydecker (commented):
Nice! A few comments on overall structure.

src/components/common/Button/Button.tsx
mbbischoff: Yeah, I was worried about that. Let’s do it.

src/components/common/Button/Button.tsx
mbbischoff: I just liked it aestethically but I think with the other change, it’ll make sense to change this too

src/components/common/Button/Button.tsx
```diff
@@ -71,35 +77,43 @@ export function Button({
...
-        />
-      )}
-      {children}
-      {EndIcon && (
-        <EndIcon
-          size={size === "large" ? 20 : 16}
-          strokeWidth={size === "large" ? 2.5 : 2}
-          {...endIconProps}
-        />
+      <span className={isSubmitting ? "invisible" : "visible"}>
```
kydecker: We need to relocate `flex` (and maybe some other styles?) to this new inner span.

CleanShot 2025-04-11 at 16 10 18@2x

CleanShot 2025-04-11 at 16 12 41@2x

mbbischoff:
Conflicts resolved

kydecker (commented):
Omg, I left this comment a week ago and forgot to hit submit review.

src/components/common/Button/Button.tsx
mbbischoff: Had to keep the flex at the top level for the vertical centering I think, but added an inline-flex to the inner span and some new storybook stories to test it.

kydecker (approved):
Tested in Storybook, everything looks great!

PR #475, opened 2025-04-11
fix: Do not nest unnecessary `Dialog` within `Popover`

kydecker:
## What changed?
- Fix `Popover` markup to allow for correct focus management when using `ComboBox`
- Remove the previously-required `title` prop from `Popover`
- Add support for placeholders in `ComboBox`.

CleanShot 2025-04-11 at 12 49 03@2x

## Why?
I encountered an issue with `ComboBox` where typing in the input would trigger the menu and then lose focus, making it impossible to type more than one character.

https://github.com/user-attachments/assets/62ee9eea-8d93-4194-bf36-03de7c8a216c

After some digging, I discovered this was because our `Popover` component had a `Dialog` nested within it. This was inserted some time ago after encountering a bug with React-Aria and warnings about missing aria-labels. It seems the bug is fixed, and manual aria-labels are no longer required (`Popover` will automatically link its trigger element via `aria-labelledby`).

With the changes to `Popover`, `ComboBox` works as expected:

https://github.com/user-attachments/assets/19233b38-4fce-47f8-9847-f9dbc886540d

## How was this change made?
- Removed inner `Dialog`
- Removed `title` prop
- Updated all components so type checking and tests pass

## How was this tested?
Re-ran `pnpm test`

## Anything else?
We aren't using `ComboBox` anywhere in the app right now, but there's a spot I want to add it soon.

### DIFF
--- a/src/components/common/ComboBox/ComboBox.tsx
+++ b/src/components/common/ComboBox/ComboBox.tsx
@@ -58,7 +58,7 @@ export function ComboBox<T extends object>({
           </FieldGroup>
           {description && <FieldDescription>{description}</FieldDescription>}
           <FieldError>{errorMessage}</FieldError>
-          <Popover title="Select an option" className="w-(--trigger-width)">
+          <Popover className="w-(--trigger-width)">
             <ListBox
               items={items}
               className="outline-0 p-1 max-h-[inherit] overflow-auto [clip-path:inset(0_0_0_0_round_.75rem)]"
--- a/src/components/common/ComboBox/ComboBox.tsx
+++ b/src/components/common/ComboBox/ComboBox.tsx
@@ -63,7 +63,7 @@ export function ComboBox<T extends object>({
           <Popover className="w-(--trigger-width)">
             <ListBox
               items={items}
-              className="outline-0 p-1 max-h-[inherit] overflow-auto [clip-path:inset(0_0_0_0_round_.75rem)]"
+              className="outline-none p-1 max-h-[inherit] overflow-auto [clip-path:inset(0_0_0_0_round_.75rem)]"
             >
               {children}
             </ListBox>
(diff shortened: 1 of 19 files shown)


src/components/common/ComboBox/ComboBox.tsx
```diff
@@ -27,6 +27,7 @@ export interface ComboBoxProps<T extends object>
   description?: string | null;
   errorMessage?: string | ((validation: ValidationResult) => string);
   children: React.ReactNode | ((item: T) => React.ReactNode);
+  placeholder?: string;
```
mbbischoff, line 30: Should we add a story and/or test for this?

src/components/common/Menu/Menu.tsx
```diff
@@ -37,14 +37,10 @@ export function MenuTrigger(props: MenuTriggerProps) {
...
-    <Popover
-      title="Select an option"
-      placement={props.placement}
-      className="min-w-[150px]"
-    >
+    <Popover placement={props.placement} className="min-w-[150px]">
       <AriaMenu
         {...props}
-        className="p-1 outline outline-0 max-h-[inherit] overflow-auto [clip-path:inset(0_0_0_0_round_.75rem)]"
+        className="p-1 outline-0 max-h-[inherit] overflow-auto [clip-path:inset(0_0_0_0_round_.75rem)]"
```
mbbischoff: sometimes we use `outline-hidden` and sometimes we use `outline-0`, should we pick one?

kydecker: `outline-hidden` [preserves the outline in forced colors mode](https://tailwindcss.com/docs/outline-style#hiding-an-outline), while `outline-0` or `outline-none` hide it everywhere.

Replacing all `outline-hidden` and `outline-0` with `outline-none`!

src/components/common/ComboBox/ComboBox.tsx
kydecker: I added an example placeholder to one of the stories! I think it's fine to leave untested, pretty small.

mbbischoff (approved):

PR #483, opened 2025-04-12
fix: Resolve Posthog console warning

### LINKED ISSUE #481: `posthog` was already loaded elsewhere. This may cause issues.
**Describe the bug**
The console reads a warning:

```
[PostHog.js] `posthog` was already loaded elsewhere. This may cause issues.
```

**To Reproduce**
1. Run the app
2. Open the browser console
3. View the warning

**Expected behavior**
Posthog is only loaded once. No warning is shown.

**Screenshots**

Image

 - OS: macOS Sequoia 15.3
 - Browser: Chrome

Ibrahimsuf:
## What changed?

Use web js to load posthog instead of `PostHogProvider`

#481 
## Why?

Posthog has a [bug](https://github.com/PostHog/posthog-js/issues/723) that causes it to load itself twice when using `PostHogProvider`.  The recommended fix was to use the web js loading instead of the `PostHogProvider`.


src/main.tsx
```diff
@@ -96,33 +99,18 @@ const InnerApp = () => {
...
 };
 
-const postHogOptions: Partial<PostHogConfig> = {
-  api_host: import.meta.env.VITE_REACT_APP_PUBLIC_POSTHOG_HOST,
-  // Since we're using TanStack Router, we need to track pageviews manually
-  capture_pageview: false,
-  // Enable web vitals
-  capture_performance: {
-    web_vitals: true,
-  },
```
kydecker: We need to keep these settings in place!

src/main.tsx
```diff
@@ -96,33 +99,18 @@ const InnerApp = () => {
...
-};
-
 const rootElement = document.getElementById("root")!;
 if (!rootElement.innerHTML) {
   const root = createRoot(rootElement);
   root.render(
     <StrictMode>
       <HelmetProvider>
         <ConvexAuthProvider client={convex}>
-          <PostHogProvider
```
kydecker, line 116: In the issue you linked there's [a comment](https://github.com/PostHog/posthog-js/issues/723#issuecomment-2636594375) that moving the provider up in the hierarchy can resolve it — can you try doing that here and see if it fixes it?

src/routes/_unauthenticated/signin.tsx
```diff
@@ -248,6 +248,7 @@ const ForgotPassword = ({
             setDidSendCode(true);
           })
           .catch((error) => {
+            console.log("Got to this step");
```
kydecker: Remove this before merge :)

src/main.tsx
```diff
@@ -14,13 +14,16 @@ import {
 import { ArrowLeft, TriangleAlert } from "lucide-react";
 import { LazyMotion, domAnimation } from "motion/react";
 import { ThemeProvider } from "next-themes";
-import type { PostHogConfig } from "posthog-js";
-import { PostHogProvider } from "posthog-js/react";
+import posthog from "posthog-js";
```
kydecker: In several other files we're currently using `import posthog from "posthog-js"`, but I think we should be using the hook `import { usePostHog } from "posthog-js/react"` instead?

.changeset/breezy-moose-remain.md
```diff
@@ -0,0 +1,5 @@
+---
+"namesake": patch
+---
+
+fix: posthog was already loaded elsewhere
```
kydecker:
```suggestion
Fix console warning about Posthog loading more than once
```
nit

kydecker (commented):
Hey @Ibrahimsuf, thanks for submitting this!

Appreciate the link to the issue filed with Posthog. Left a few comments for other things to try.

src/main.tsx
Ibrahimsuf: You are correct if we move `PostHogProvider` to to the top it fixes the warning.

src/main.tsx
Ibrahimsuf: It seems like we can use `PostHogProvider` as long as it is high enough in the hierarchy.

src/main.tsx
Ibrahimsuf: I don't know what exactly it needs to be ahead of, but it only works at the top of the `root.render` block.

src/main.tsx
kydecker: Right, I mean anywhere else in the application that currently has `import posthog from "posthog-js"` needs to be replaced with a hook.

[From the docs:](https://posthog.com/docs/libraries/react#using-posthog-js-functions)

> Do not directly import posthog apart from installation as shown above. This will likely cause errors as the library might not be initialized yet. Initialization is handled automatically when you use the PostHogProvider and hook.

src/main.tsx
kydecker: We still need to keep it inside of `StrictMode`. Try relocating it there and replacing other posthog imports with the `usePosthog` hook.

If that doesn't fix it, we may also need to [initialize the client instance](https://posthog.com/docs/libraries/react#posthog-provider) instead of passing in `apiKey` and `options`.

kydecker:
@Ibrahimsuf I took the time to get the rest of this PR over the finish line! Thank you for your contribution!

kydecker (approved):

PR #496, opened 2025-04-23
chore: Remove references / urls from `quests` schema

kydecker:
## What changed?
Followup to #493. Remove `references`/`urls` from `quests` schema.

## Why?
Expanding the editor component to support more content means that our schema can be less rigid. In other words, instead of putting references in a dedicated column, we can just put them in the body of the quest contents.

## How was this change made?
Deleted `urls` column from the `quests` table and all related models.

## How was this tested?
N/A. Existing tests pass.

### DIFF
--- a/convex/model/questsModel.ts
+++ b/convex/model/questsModel.ts
@@ -174,63 +174,6 @@ export async function update(
   });
 }
 
-export async function updateUrls(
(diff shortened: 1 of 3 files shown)


mbbischoff (commented):
lgtm

Issue #480, opened 2025-04-11
Remove questSteps table and content


evadecker:
Once #479 is merged, we should delete the `questSteps` table and migrate any existing content. In production right now it may be easier to just delete the content and recreate it; there isn't much there yet anyways.

Delete any components related to `questSteps` and delete the `Steps` component which currently exists outside of the `Editor`.

evadecker:
Resolved in #497

PR #500, opened 2025-04-25
fix: Editor `onUpdate` events

kydecker:
## What changed?
#480 introduced `QuestContent` but did not pass values to `onUpdate`. Fix.


mbbischoff:
✅

Issue #502, opened 2025-04-25
Batch quest page updates and save all at once


evadecker:
Currently, individual elements of the quest are saved individually: costs, title, time required, content...

In order to facilitate history for the fields and improve UX, there should be a single "Save changes" button which saves *all* the values of the quest edit form.

- [ ] If the form is dirty, it should [block navigation](https://tanstack.com/router/latest/docs/framework/react/guide/navigation-blocking) without confirmation
- [ ] Form edits should be saved in localStorage before submission in order to prevent data loss and allow resuming if the page is closed
- [ ] A library or helper context may need to be introduced to manage state for all the various form elements
- [ ] A badge should display clearly at the top of the page indicating the save status: empty (no changes), "unsaved changes" (form is dirty), or "saved" (display temporarily after save).

PR #503, opened 2025-04-25
fix: Add "save changes" button to `QuestContent` editor

kydecker:
## What changed?
Add an explicit button for saving contents of a quest, until we can implement #502.


mbbischoff:
Makes sense, code looks good

PR #508, opened 2025-04-26
fix: Fix sign in and sign out redirects

### LINKED ISSUE #277: Reset password flow and sign in hangs forever, requires refresh to view the right page
**Describe the bug**
When resetting a user password, after entering the code and new password, the button enters a disabled state the the screen does not reload. After a hard refresh, the user will be logged in, but it does not happen automatically.

**To Reproduce**
1. Go to /login
2. Click on "Forgot?"
3. Enter the code sent to the registered email along with the new password
4. Button will disable, page will not reload.

**Expected behavior**
Page reloads / refreshes once password is reset and user is logged in.

 - OS: macOS
 - Browser: Chrome
 - Version: 130

kydecker:
## What changed?
Fixes #277 !!!

Adds a Promise which awaits a query call to confirm that a user successfully logged in, then redirects.

## Why?
Previously, signIn() and signOut() were redirecting before Convex had fully authenticated a user, resulting in a failed navigation attempt. This seems like one of the rough edges with Convex Auth right now. Thanks to help from the Convex Discord, we can fix it by adding a Promise which waits for an auth confirmation from the backend before redirecting the user.

## How was this change made?
Created a new `waitFor()` utility function which sets and checks the results of a Promise. Used that to await a successful backend call to Convex before redirecting on sign in or sign out.

Changed previous calls to `signOut()` to instead navigate to a new `/signout` route which handles the above logic.

## How was this tested?
Manually confirms:

- [x] Redirects on sign in
- [x] Redirects on password reset + code confirmation
- [x] Redirects on sign out
- [x] Redirects when attempting to visit `/signin` and already authenticated
- [x] Redirects to `/signin` when attempting to access authenticated routes and not signed in

### DIFF
--- a/.changeset/six-shrimps-occur.md
+++ b/.changeset/six-shrimps-occur.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Fix signin and signout redirects
--- a/convex/users.ts
+++ b/convex/users.ts
@@ -99,3 +99,7 @@ export const deleteCurrentUser = userMutation({
     await Users.deleteUser(ctx, { userId: ctx.userId });
   },
 });
+
+export const isSignedIn = query(async (ctx) => {
+  return !!(await getAuthUserId(ctx));
+});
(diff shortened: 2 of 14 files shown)


mbbischoff:
> This seems like one of the rough edges with Convex Auth right now. Thanks to help from the Convex Discord, we can fix it by adding a Promise which waits for an auth confirmation from the backend before redirecting the user.

Did they mention what they see doing about this longer term?

kydecker:
> Did they mention what they see doing about this longer term?

Yep, they just shared a Convex roadmap which included improvements to Convex Auth and/or relying more on Better Auth.

Either way it sounds like this should get easier going forward and we can remove the extra check. But this fix was shared by one of the engineers who writes the auth library.

PR #515, opened 2025-04-27
chore: Exclude routes from coverage testing, add thresholds for coverage

kydecker:
## What changed?
- Exclude everything in `src/routes` from coverage testing (at the route-level it's easier and probably more important to use e2e/integration tests instead)
- Add minimum thresholds for coverage
- Separately, remove the `ReadingScore` component and dependencies (old exploration)

## Why?
Ensure that new PRs do not cause coverage to lower. More accurately represent the % of components and utils that are tested.


mbbischoff:
Sweet, makes sense

PR #520, opened 2025-04-28
test: Add Playwright tests for sign-in and registration

### LINKED ISSUE #297: Create new Playwright tests to validate sign in and registration
Create a new test within `tests/e2e` to test the end-to-end flow for signup, registration, and password reset.

### Registration
- Displays an error message displays when fields are invalid; form does not submit
- Displays a success message when submitting a valid registration
- Alerts the user if the email address for registration already exists (requires #294)

### Password reset
- Displays an error message if the account is not found
- Displays the code entry success message and input if the code was sent
- Allows entering the emailed code and resetting the password (will require mocking)

### Sign in
- Displays an error message if email does not belong to a registered user (requires #295)
- Displays error message if the email or password fields are invalid (malformatted email or empty)
- Logs the user in and redirects after successful submission

Because we haven't written many Playwright tests for Namesake yet, when tackling this ticket it may be necessary to create additional test helpers or configuration in order to simplify the test writing process. If it seems likely that other Playwright tests will require similar mocks or helpers, consider abstracting them somewhere reusable.

kydecker:
## What changed?
Add Playwright end-to-end tests for sign in and registration. Resolves #297 

## Why?
Maintain confidence that all of our core flows are stable!

## How was this change made?
Added tests for some basic functionality.

## How was this tested?
They are tests.

## Anything else?
Fix a few accessibility errors on the signin page.

Will probably need to come back and abstract out some of the signin logic to reuse it for other Playwright tests.


playwright.config.ts
```diff
@@ -13,16 +13,20 @@ export default defineConfig({
...
+      use: { ...devices["Desktop Firefox"] },
+    },
     {
       name: "Mobile Safari",
       use: { ...devices["iPhone 12"] },
     },
   ],
   webServer: {
-    command: "pnpm dev",
+    command: "npm run dev",
```
mbbischoff: Why the change from pnpm to npm here?

mbbischoff:
GitHub actions seems unable to start the server

> [WebServer] ✖ Cannot prompt for input in non-interactive terminals.
> 
> Error: Process from config.webServer was not able to start. Exit code: 1

mbbischoff (approved):
Tests look good once they can run okay

playwright.config.ts
kydecker: To speed up Playwright tests this relies on Microsoft's official Docker container for Playwright. This speeds up CI a ton but it doesn't have pnpm installed so we have to use npm. Prob worth adding a comment in the file!

kydecker:
Thanks! I remember running into a block here but I can't remember what it was. Will finish up this PR soon

PR #554, opened 2025-05-11
feat: Show signup screen for first time users

### LINKED ISSUE #532: Default to "Register" tap on /signin if user has not visited / signed in before
When users visit for the very first time, we should default to the "Register" tab instead of "Sign in". For everything else we should display the "Sign in" tab.

mbbischoff:
## What changed?
Fixes #532.

## Why?
Because first time visitors probably want to sign up.

## How was this change made?
* Added local storage bool that’s set on first visit in an actual browser.

## How was this tested?
* Visited the site and then visited it again.

## Anything else?
This could use tests, so I’m marking this as a draft but if you like the approach here @evadecker, I can add them before this goes in.


mbbischoff:
@evadecker Can I get your thoughts here?

kydecker:
> @evadecker Can I get your thoughts here?

Sorry for the delay! Simple and straightforward; makes sense to me.

mbbischoff:
@evadecker do have thoughts on the right way to test this?

mbbischoff:
Okay added some unit tests here

kydecker (approved):
LGTM!

If you want we can relocate this function and test into `/hooks` and call it with `useHasVisitedBefore()`

But current approach also works great, up to you

PR #566, opened 2025-05-21
content: Add Social Security guided form

kydecker:
## What changed?
Adds PDF definitions, tests, and UI for filling out the Social Security form (SS-5).

## Why?
Trying to get this in before the bug bash so we can test!

## How was this change made?
- Renamed editable form fields within the `.pdf`
- Wrote `.ts` field mappings
- Added tests to verify PDF gets filled properly
- Added UI and a new `/forms/social-security` route
- Added constants for all countries including their ISO two-letter country codes

## How was this tested?
- Added tests for the pdf form filling
- Added a test for a new pdf helper used to display birth state or country
- Manually filled out fields and clicked save and download to test that the PDF downloads and displays correctly

## Anything else?
This is the second full guided form developed for the site! The flow feels pretty good, although there may be additional opportunities to cut down on the amount of boilerplate. All-in-all I feel pretty happy with how easy this was to set up. Most of the time was spent renaming existing editable fields on the PDF.


convex/_generated/api.d.ts
```diff
@@ -165,82 +165,5 @@ export declare const components: {
         }
       >;
     };
-    public: {
```
kydecker: Not sure why this changed. Seems like an automatic Convex thing

mbbischoff:
Reviewing now

src/constants/countries.ts
```diff
@@ -0,0 +1,250 @@
...
+  VN: "Viet Nam",
+  VG: "Virgin Islands, British",
+  VI: "Virgin Islands, U.S.",
+  WF: "Wallis and Futuna",
+  EH: "Western Sahara",
+  YE: "Yemen",
+  ZM: "Zambia",
+  ZW: "Zimbabwe",
+};
+export type Country = keyof typeof COUNTRIES;
```
mbbischoff, line 250: Should we maybe use a library for this in the future so it stays up to date?

src/routes/_authenticated/forms/social-security.tsx
```diff
@@ -0,0 +1,339 @@
...
+              Social Security Administration already has on file.
+            </strong>{" "}
+            The gender marker is not shown on your Social Security card.
+          </p>
+          <p>
+            As of January 31, 2025, the Trump administration has directed the
+            Social Security Administration to stop processing gender marker
+            updates associated with social security records. If you apply for a
+            gender marker change with Social Security now, you will likely be
+            told that they cannot process your request.
```
mbbischoff: Can we link out here if folks want to read more?

mbbischoff (approved):
Looks great, mostly just thoughts for small improvements

src/constants/countries.ts
kydecker: I initially installed https://github.com/fannarsh/country-list but it came with boilerplate we don't need... felt easier to cut out the dependency since countries and their codes aren't changing very often.

src/routes/_authenticated/forms/social-security.tsx
kydecker: Added!

PR #573, opened 2025-05-21
fix: Update link logic to remove `useHref` and pre-populate internal links

### LINKED ISSUE #527: Button within editor is not working consistently, should allow internal links
Right now the `Button` component used within the Editor only allows full URLs, so it's not possible to insert a relative link.

Image

- Add support for internal links which adhere to `ToOptions` in TanStack Router, e.g `{ to: 'forms/ma-court-order' }`
- Verify that the button link rendered in non-edit mode is clickable locally and in production (Button is currently not working as intended)
- Update the popover to add tabs between internal and external links. Default to internal.
- Update tests.

kydecker:
## What changed?
Followup to #567 and #571. Remove `useHref` when setting links because we are passing in the full `ToOptions` object instead of just `to`.

Updates the popover form for setting links within a quest. Resolves #527.

CleanShot 2025-05-21 at 17 36 37@2x
CleanShot 2025-05-21 at 17 36 42@2x

## Why?
The links within quests to forms were not working correctly, and the UX for actually setting the links was poor because it required a full URL instead of providing links to internal routes.

## How was this tested?
Updated unit tests for `ButtonComponent` to accommodate new logic.
Manually tested changing links and viewing content.

## Anything else?
Modified the size of the inputs and the tooltip text to change "URL" to "link".

### DIFF
--- a/src/components/common/Editor/extensions/ButtonComponent.tsx
+++ b/src/components/common/Editor/extensions/ButtonComponent.tsx
@@ -24,7 +24,7 @@ import { LinkIcon, Unlink } from "lucide-react";
 import { useRef, useState } from "react";
 import { useInteractOutside } from "react-aria";
 
-type LinkType = "namesake" | "external";
+type LinkType = "internal" | "external";
 
 export default function ButtonComponent({ editor, node }: NodeViewProps) {
   const { routesByPath } = useRouter();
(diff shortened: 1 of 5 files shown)


src/components/common/Editor/extensions/ButtonComponent.tsx
```diff
@@ -86,22 +119,58 @@ export default function ButtonComponent({ editor, node }: NodeViewProps) {
...
+                </TabList>
+                <TabPanel
+                  id="internal"
+                  className="flex flex-row items-start gap-1 w-full"
+                >
+                  <Select
+                    aria-label="Select path"
+                    selectedKey={form}
+                    onSelectionChange={(value) => {
+                      setForm(value.toString());
```
mbbischoff: Is the check failure here correct that value could be null?

mbbischoff (approved):

src/components/common/Editor/extensions/ButtonComponent.tsx
kydecker: Yeah just fixed

PR #588, opened 2025-05-23
feat: Replace `Select` elements with `ComboBox`

### LINKED ISSUE #582: Replace some `Select` components with `ComboBox`
https://namesakefyi.github.io/namesake/?path=/story/components-common-combobox--example

For fields like the state selection in settings, it would be better to use a `ComboBox` instead of `Select` in order to support easier filtering of list items.

It may also be worth changing the following:

- Language select
- Country select
- Birthplace select
- Residence select
- Any other long select dropdowns

### LINKED ISSUE #584: Some settings fields cramped on mobile
Image

Explore ways to prevent settings items from getting squished on mobile

kydecker:
## What changed?
Resolves #582. Replaces several `Select` elements within forms and user settings with a `ComboBox`, to make it easier to filter and select items.

https://github.com/user-attachments/assets/e3f0236f-e427-483b-89c7-5ccca4aa9137

https://github.com/user-attachments/assets/f13ea155-a748-4821-95b2-b15a24c89874

https://github.com/user-attachments/assets/8793a5e5-bc71-42d7-933e-480d4668e727

## Why?
`ComboBox` offers better support for keyboard navigation because it allows users to type and filter a list of results inline.

## How was this change made?
- Replaced `Select` with `ComboBox` for `EditBirthplaceSetting` and `EditResidenceSetting`, updated tests
- Created a new `ComboBoxField` component for use within guided forms
- Replaced UI for `LanguageSelectField`, for the state and county selection within `AddressField`, and for the state and country selection when asking for birthplace
- Installed `react-aria/test-utils` to assist with managing ComboBox state in unit tests, see: https://react-spectrum.adobe.com/react-aria/testing.html

## How was this tested?
Updated existing tests with new cases; wrote new tests where components were created.

## Anything else?
- Slightly shrinks the size of the icon for `large` size `Button`
- Adds support for `size` to `ComboBox` using the same styling and logic as other components like `Select`
- Passes down `autoComplete` and `name` props from `ComboBox` to its internal `Input` field
- Changed "I was born outside the US" to "Outside the US" for brevity and applies truncation to the settings button. Resolves #584


kydecker:
After recording the videos in the description, I decided to prevent the popover from flipping sides; felt jarring to have it pop up on top for a long list and then flip to the bottom for a small one. It'll always open on the bottom now

mbbischoff:
@evadecker This wasn’t exactly what I expected the fix to be. I think I was expecting that we’d be able to have the dropdowns allow typeahad like the standard controls browsers provide do? 

Testing on https://react-spectrum.adobe.com/react-aria/Select.html, it looks like React-ARIA does support that so I wonder if there’s something we’re doing to prevent it. 

---

Also, if you prefer this UX, I’m happy to just say let’s go with this instead.

kydecker:
@mbbischoff Yeah typing in a Select field *should* work! (It does for me?) https://namesakefyi.github.io/namesake/?path=/story/components-common-select--example

But overall I do think ComboBox is a better fit here, a lot of people don't know that you can type to pick inside Select.

mbbischoff:
> @mbbischoff Yeah typing in a Select field _should_ work! (It does for me?) https://namesakefyi.github.io/namesake/?path=/story/components-common-select--example
> 
> But overall I do think ComboBox is a better fit here, a lot of people don't know that you can type to pick inside Select.

Good point okay cool

kydecker:
@mbbischoff Going back on what I said earlier! Thought about it some more and I think the UX for ComboBox within settings is funky. Popovers inside popovers. Reverted back to `Select`. Kept the form questions as `ComboBox`, though.

PR #595, opened 2025-05-25
feat: Make it easier to browse all available quests

kydecker:
## What changed?
Implement a tab view within `QuestsNav` to make it easier for non-admin users to browse and preview all available quests.

https://github.com/user-attachments/assets/bdab4f57-6da8-4231-95f5-6f3bed788f4e

## Why?
As new quests get added people will need to search and add ones. Rather than dedicate an entirely separate view to browsing, I thought it made sense to reuse the existing `QuestsNav` sidebar, which can use the same navigation patterns on mobile and desktop, and make it easy to quickly preview each quest.

Would love design feedback, though!

## How was this change made?
- Updated `Tabs` component to support a new `size` prop
- Updated `QuestsNav` with the new tabs for "My quests" and "All quests"
- Created new "All quests" view which groups available quests by category, displays them in nav groups, and displays an "add" button if the quest is not already in the user's list
- Updated `StatusSelect` to support the "condensed" view (icon-only) and updated the nav items to support changing status directly from the nav instead of needing to click in
- Changed "my list" copy to "my quests" throughout the app

## How was this tested?
Manual for now; will update tests once design is locked down.

## Anything else?
- We will need to improve search and filtering at some point. Super basic implementation for now.
- Updated the design and fixed some issues with the `GridList` component; originally I intended to use `GridList` for this new browse screen, but it didn't end up being necessary; still, I wanted to include the fixes to the component


kydecker:
@mbbischoff Curious to get your thoughts on design here!

kydecker:
@mbbischoff Gonna merge this for now but down to iterate and ship further improvements 💕

PR #606, opened 2025-05-30
fix: Add support for "prefers not to answer" in form fields

### LINKED ISSUE #579: Currently impossible to deselect radio options
Feedback from MG: Social Security Card question on identifying as Hispanic or Latino and race: add option "Decline to State" since it is optional.

Currently, if a user selects "yes" or "no" there is no way to deselect the option.

We should allow deselecting a radio option by clicking the same option again.

Additionally, it may be worth considering adding explicit "Decline to answer" options for questions which are optional.

kydecker:
## What changed?
Allow users to opt out of answering certain questions. Resolves #579.

https://github.com/user-attachments/assets/45f715ba-aa40-4139-bf4b-a151561bdc30

## Why?
Certain form questions are optional, but radio groups cannot be deselected, and checkbox groups require some extra work to uncheck and disable all the other checkboxes.

## How was this change made?
1. Add a new `includePreferNotToAnswer` boolean prop for `CheckboxGroupField`, `RadioGroupField`, and `YesNoField`. Use the same naming and labeling scheme, and always include the item at the end of the list
2. For `CheckboxGroupField`, disable and deselect all other checkbox items
3. Update tests
4. Update the social security form questions for ethnicity and race to include "Prefer not to answer"

## How was this tested?
Wrote new unit tests and updated existing ones. Tested manually.

## Anything else?
Dove into the history of radio buttons on the web and why they are not de-selectable. Considered adding some JS to allow click-to-disable on radio buttons, but ultimately felt that (at least from a user psych perspective) most people will want to answer every question in a form, even if their answer is "I don't want to answer". This shouldn't affect any backend logic. Verified that "prefer not to answer" responses save and display correctly within the user responses view, and don't crash creating a PDF.

### DIFF
--- a/.changeset/slimy-mangos-grab.md
+++ b/.changeset/slimy-mangos-grab.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Support selecting 'prefer not to answer' for certain optional questions
--- a/src/components/forms/CheckboxGroupField/CheckboxGroupField.test.tsx
+++ b/src/components/forms/CheckboxGroupField/CheckboxGroupField.test.tsx
@@ -85,4 +85,59 @@ describe("CheckboxGroupField", () => {
     const childComponent = screen.getByTestId("child-component");
     expect(childComponent).toBeInTheDocument();
   });
+
+  it("renders prefer not to answer option when enabled", () => {
+    renderWithFormProvider(
+      <CheckboxGroupField
+        name="reasonForChangingName"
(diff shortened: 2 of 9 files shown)


mbbischoff (approved):
Code looks good. We’re only using this on fields that are actually optional on the governement form, right?

kydecker:
> Code looks good. We’re only using this on fields that are actually optional on the governement form, right?

That's right

Issue #530, opened 2025-04-29
Send welcome/onboarding email on registration


evadecker:
Send an email to users who register for Namesake.

Welcome them and explain the process + how the site works.

Have a button which links to the Discord.

Have other links for contact + support.

Issue #286, opened 2024-12-17
Verify email addresses


evadecker:
Implement email verification during registration in order to combat spam, abuse, etc.

Email verification should also trigger when the user changes their email in settings.

A user should be able to log in and use the app without verifying, but we should display persistent notifications in settings and on their home page to encourage verification.

Follow the Convex Auth guide for setting up email verification using Resend: https://labs.convex.dev/auth/config/passwords#email-verification-setup

mbbischoff:
Started on https://github.com/namesakefyi/namesake/tree/email-verification

Issue #76, opened 2024-09-18
Implement `emailVerified` logic for `users` table


evadecker:
#74 relocated some account creation logic and no longer sets an `emailVerified` timestamp or boolean. Fix this.

PR #609, opened 2025-05-31
feat: Use Better Auth and improve Namesake emails

kydecker:
## What changed?
Switch from Convex Auth to [Better Auth](https://www.better-auth.com/). Unblocks #76 and #286.

- Delete account now sends a confirmation email + has an updated confirmation dialog

https://github.com/user-attachments/assets/12eeabdc-c218-44b7-97db-04c3baf4f62b

- Registration form now asks for display name and saves it to the user's "name" field

CleanShot 2025-06-01 at 13 40 05@2x

- Created and styled new emails using [React Email](https://react.email/)! (Preview by running `pnpm dev:email` or `pnpm email dev` and opening localhost:3000)

| Email | Preview |
|--------|--------|
| Account deletion | CleanShot 2025-06-01 at 13 36 16@2x |
| Password reset | CleanShot 2025-06-01 at 13 37 13@2x |
| Welcome (unused, ticketed to implement in #530) | CleanShot 2025-06-01 at 13 37 33@2x | 

- Reset password flow now includes a [Sniper Link](https://sniperl.ink/) to automatically display a button linking to the user's email client and defaulting to a filtered inbox view.

https://github.com/user-attachments/assets/a4727bb5-85fd-4fd4-937d-4e5bbd519276

## Why?
Better Auth is more mature and provides a much more robust framework for dealing with user identities. Convex is investing in an official Better Auth integration, and it makes sense for us to move to it before launch.

## How was this change made?
- Implemented the [Convex + Better Auth integration](https://convex-better-auth.netlify.app/#getting-started) following the guide
- Removed Convex Auth providers and packages
- Installed [react-email](https://react.email/) to assist with writing styled email messages
- Verified that emails send and display properly on desktop and mobile
- Split apart the `/signin` route into different components for sign in, registration, forgot password, and reset password; created new routes for `/forgot` and `reset-password`
- Implemented forgot password and reset logic following the [Better Auth guide](https://www.better-auth.com/docs/authentication/email-password#forget-password)
- Verified that passwords can be properly reset via the email confirmation flow
- Added [Sniper Link](https://sniperl.ink/) to provide a button to open the user's mail client

## How was this tested?
Wrote new unit tests for new components. Manually sign up, registration, password reset, and account deletion. Will update #520 to use Better Auth implementation for E2E tests once this is merged.

## Anything else?
- Added `aria-label="Loading"` to `Button` when `isSubmitting` is true.
- Merging this as-is will require everyone who is signed up to create a new account; there's no migration plan between Convex Auth and Better Auth, which would require a lengthy implementation where both auth systems are active in order to passively accept users' passwords and hash them within Better Auth. (We can't migrate hashed passwords from one provider to another.) Since so few people have accounts (less than 10) I think it makes sense to just kill Convex Auth and let people know to re-register. Their user data will still be there.
- I updated the contribution manual but I haven't gone through the process myself, so there might be extra steps we need to get contributors set up using Better Auth.
- Note that there are two copies of the image assets for emails: one within `emails/static` (used for testing React Email locally) and one within `public/email` (which will be hosted and fetched once the PR is merged). If you view the "live" emails in your inbox right now, the images will be broken, because they aren't hosted yet; this should be fixed after merge.


mbbischoff:
This is a big PR and I want to spend time testing it, so it may take me a bit longer.

mbbischoff:
CleanShot 2025-06-06 at 16 27 12@2x

Can we allow 2 letter display names? Asking for me

mbbischoff:
Not sure why but I’m getting a console error when trying to register on this branch:

> Fetch API cannot load https://modest-caterpillar-513.convex.site/api/auth/sign-up/email due to access control checks.

Did something change with convex access?

mbbischoff:
Looks like maybe we need to run the email bodies through the smart quote function.

CONTRIBUTING.md
```diff
@@ -43,14 +43,14 @@ Once your Convex account and project is created, a `.env.local` file will be gen
 
 ### Set up authentication (first time only)
 
-Namesake authenticates users with [Convex Auth](https://labs.convex.dev/auth). To initialize authentication for local development, run:
+Namesake authenticates users with [Better Auth](https://convex-better-auth.netlify.app/). To connect, you will need to add a local environment variable. Run the following to generate a secret. 
 
 ```shell
-npx @convex-dev/auth
+npx convex env set BETTER_AUTH_SECRET=$(openssl rand -base64 32)
 ```
```
mbbischoff: Followed these steps but still hit a 404 in the console when trying to register locally:

> Fetch API cannot load https://modest-caterpillar-513.convex.site/api/auth/sign-up/email due to access control checks.

convex/auth.ts
```diff
@@ -1,15 +1,112 @@
...
+      // Delete userSettings
+      const userSettings = await ctx.db
+        .query("userSettings")
+        .withIndex("userId", (q) => q.eq("userId", userId as Id<"users">))
+        .first();
+      if (userSettings) await ctx.db.delete(userSettings._id);
+
+      // Finally, delete the user
+      await ctx.db.delete(userId as Id<"users">);
     },
```
mbbischoff: From a separation of concerns perspective, might it make sene to have this code live closer to the user model itself rather than in auth and then auth could just call functions exposed by user?

emails/_components/EmailFooter.tsx
```diff
@@ -0,0 +1,80 @@
...
+  Img,
+  Link,
+  Row,
+  Section,
+  Text,
+} from "@react-email/components";
+import { baseUrl } from "../constants";
+import { styles } from "../styles";
+
+export const EmailFooter = () => {
```
mbbischoff, line 13: Lovely design

mbbischoff:
Looks like the email graphics live in two places, why is that?

src/components/common/Button/Button.tsx
```diff
@@ -108,6 +108,7 @@ export function Button({
   return (
     <AriaButton
       {...props}
+      aria-label={isSubmitting ? "Loading" : props["aria-label"]}
```
mbbischoff, line 109: Thanks, good catch

mbbischoff (changes requested):

kydecker:
> CleanShot 2025-06-06 at 16 27 12@2x
> 
> Can we allow 2 letter display names? Asking for me

Oof! Good call. Can't believe I got caught by one of the [falsehoods](https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/)...

convex/auth.ts
kydecker: That's a great suggestion, will update

CONTRIBUTING.md
kydecker: Ah. This is due to the `baseUrl` that's set in `main.tsx`.

Right now it's hardcoded as https://modest-caterpillar-513.convex.site, but that won't be the correct URL for contributors, since Convex will generate a unique URL for your local dev environment.

This is also causing problems with setting `redirectTo`, which I [asked about in Discord](https://discord.com/channels/1019350475847499849/1378458750091788420)...

CleanShot 2025-06-06 at 18 19 11@2x

CleanShot 2025-06-06 at 18 19 32@2x

If you want to test locally, I think you can replace it with your own dev URL:

https://discord.com/channels/1019350475847499849/1378458750091788420

But hopefully there will be a better solution here soon.

kydecker: Confirmed that we should update `baseUrl` here to reference `.env` or `.env.local`

convex/auth.ts
kydecker: Relocated into `authModel`

CONTRIBUTING.md
kydecker: This should work now as long as you have this defined in `.env.local`:

```
VITE_CONVEX_SITE_URL=https://adjective-animal-123.convex.site
```

Replacing `adjective-animal-123` with whatever yours is. Note this should be the same as the `.convex.cloud` url but with `.site` instead of `.cloud`

kydecker:
> Looks like the email graphics live in two places, why is that?

When testing emails in the dev server it requires images to be in `/emails/static` *or* hosted publicly via CDN, etc. Since the images aren't deployed/hosted yet, they need to go in `/emails/static` for local testing and `/public` for the final deploy.

An alternative could be pre-uploading email image assets during dev but ¯\_(ツ)_/¯ this felt easier

kydecker:
> Looks like maybe we need to run the email bodies through the smart quote function.

This is difficult to do automatically due to the structure of the React nodes within the email, but I've gone ahead and replaced all the straight quotes with curly quotes.

mbbischoff:
Re-reviewing now

CONTRIBUTING.md
mbbischoff: I believe from tiral and error that it’s also important to set `VITE_CONVEX_SITE_URL`

mbbischoff (approved):
Code looked good last time I read it over. Tested it this morning and it worked the way I expected. Just added one note on an extra step I think is required for the CONTRIBUTING.

CONTRIBUTING.md
kydecker: Updated https://github.com/namesakefyi/namesake/pull/609/commits/aa37d8dc413fb7ba0074917ee44418761467a554

PR #610, opened 2025-06-01
fix: Add query and mutation safeguards for admins

### LINKED ISSUE #526: Convert select backend queries and mutations to `internalMutation`, `internalQuery` or gate by `admin` role
We need to implement safeguards in our Convex queries against anyone calling e.g. `getAll()` on `users.ts`.

- For queries and mutations which are only called internally, define them as `internalMutation()` `internalQuery()`.
- For queries and mutations which should only be called by admins, add a role check.
- Update tests to verify that endpoints are only callable where intended.

kydecker:
## What changed?
Fixes #526.

## Why?
Plain Convex `query()` and `mutation()` are publicly-accessible and have the potential to expose, for example, a list of all users. Bad!

This adds role checking for several queries and mutations to ensure that only admin roles can run them.

## How was this change made?
- Created a new `adminQuery` custom query and `adminMutation` custom mutation which can be imported and used.
- Added new helper functions for `createAdmin()` and `createUser()` for re-use in tests

## How was this tested?
Updated tests to run `asAdmin` where necessary.

## Anything else?
Improve some test coverage. Since function coverage is now above 80%, increase threshold to 80 to match other coverage targets.


mbbischoff:
Trusting the test suite on this one but will review the code.

convex/model/userSettingsModel.ts
```diff
@@ -9,7 +9,7 @@ export async function getSettingsForUser(
   const userSettings = await ctx.db
     .query("userSettings")
     .withIndex("userId", (q) => q.eq("userId", userId))
-    .first();
+    .unique();
```
mbbischoff, line 12: Why the change from first to unique?

convex/helpers.ts
```diff
@@ -23,3 +25,65 @@ export const userMutation = customMutation(mutation, {
...
+  const adminId = await t.run(async (ctx) => {
+    return await ctx.db.insert("users", {
+      email: email ?? "admin@namesake.fyi",
+      name: name ?? "Test Admin",
+      role: "admin",
+    });
+  });
+  const asAdmin = t.withIdentity({ subject: adminId });
+  return { adminId, asAdmin };
+};
```
mbbischoff: I think making `email` and `name` both optional as a way to signal creating the test admin might get someone into trouble down the road if they don’t read this implementation and pass just an email not knowing that it’ll assign that user a name of Test Admin or vice-versa.

convex/helpers.ts
```diff
@@ -23,3 +25,65 @@ export const userMutation = customMutation(mutation, {
...
+  const userId = await t.run(async (ctx) => {
+    return await ctx.db.insert("users", {
+      email: email ?? "user@example.com",
+      name: name ?? "Test User",
+      role: "user",
+    });
+  });
+  const asUser = t.withIdentity({ subject: userId });
+  return { userId, asUser };
+};
```
mbbischoff: See comment below.

mbbischoff (approved):
Love all the deletions of boilerplate in here. Minor notes.

convex/model/userSettingsModel.ts
kydecker: `.first()` will always return if data exists; `.unique()` is a little more strict and will throw an error if the query returns more than one result.

Since there shouldn't be multiple row entries for the same userId I figured it made sense; plus `.unique()` seems to be used more often than `.first()` in Convex docs

convex/helpers.ts
kydecker: Hm, what would you recommend?

Within the database `email` and `name` are both required, but within tests I wanted to keep the call site as minimal as possible so it's possible to just say `createAdmin()` and get a test admin user back.

I could update the props here with JSDoc comments to explain, so at least people get Intellisense? And maybe rename this to `createTestAdmin()`...

convex/helpers.ts
kydecker: Ditto. WDYT about adding JSDoc comments and renaming this to `createTestUser`

mbbischoff: yeah, i think that solves it well

kydecker: Updated names and added docs!

CleanShot 2025-06-06 at 22 05 43@2x

CleanShot 2025-06-06 at 22 04 25@2x

Also relocated these two functions to a new `__test__/helpers.ts` file since it felt like they should be separate from `convex/helpers.ts`, which contains functions for prod data

PR #619, opened 2025-06-03
feat: Add cover page to PDF download packets

### LINKED ISSUE #580: Add cover page for PDF packet downloads
Each PDF packet download should include a cover page which contains:

1. Instructions for what to do with the document
2. A list of all the documents contained in the packet
3. Fields that must be filled in by hand
4. Date created and Namesake logo

These items should be dynamically generated using pdf-lib.

kydecker:
## What changed?
Resolves #580.

| Court Order | Social Security |
|--------|--------|
| CleanShot 2025-06-03 at 12 37 20@2x | CleanShot 2025-06-03 at 12 38 01@2x | 

## Why?
- Provide a place to instruct users on what to do with their form
- Act as a privacy sheet when printing documents on a public printer (library, etc.)
- Legal disclaimer

## How was this change made?
Created new `createCoverPage()` form utility function. Integrated it inside `downloadMergedPdf()` function.

Allows passing in a title and a list of instructions.

All documents are dynamically listed by their full title and code.

## How was this tested?
Hard to test PDFs but wrote a very basic unit test. Probably better to end-to-end test this.

## Anything else?
We could improve these at some point with more details about where to go and what to do. (Like: based on your address, here's your courthouse!)


kydecker:
Design could probably be polished a bit more, will return to this tomorrow

kydecker:
Designs updated!

Unfortunately I can't find a way to dynamically calculate the height of text areas that get inserted onto the cover page, which means we have to set an absolute value for each section to appear (instead of relatively positioning everything with consistent margins between them).

I tried to pick spots that leave enough room for title (3 lines), documents (up to 7) and instructions (all the remaining space).

src/forms/utils.ts
```diff
@@ -154,3 +181,143 @@ export async function downloadMergedPdf({
     console.error(error);
   }
 }
+
+/**
+ * Create a cover page PDF for a document packet.
+ */
+export async function createCoverPage({
```
mbbischoff, line 188: There’s a lot of magic numbers in here, but I don’t really see a ton of value in making them constants right now. But maybe down the road we could have a better system for this layout and this could get easier to read.

src/forms/utils.ts
```diff
@@ -154,3 +181,143 @@ export async function downloadMergedPdf({
...
+  }
+
+  const currentDate = new Date().toLocaleDateString("en-US", {
+    month: "long",
+    day: "numeric",
+    year: "numeric",
+    hour: "2-digit",
+    minute: "2-digit",
+  });
+  const generatedOn = `Document generated from https://app.namesake.fyi on ${currentDate}.`;
```
mbbischoff:
```suggestion
  const generatedOn = `Document generated by https://app.namesake.fyi on ${currentDate}.`;
```

src/routes/_authenticated/forms/social-security.tsx
```diff
@@ -81,6 +81,16 @@ function RouteComponent() {
 
       await downloadMergedPdf({
         title: "Social Security Card",
+        instructions: [
+          "Please review all documents carefully.",
+          "Fill in your social security number—for security, Namesake never asks for this.",
```
mbbischoff, line 86: Love this

mbbischoff (approved):
Design is nice :)

src/forms/utils.ts
kydecker: Yeah :/ Def could be clearer. I wish it was possible to lay out a page using CSS or something. The methods for PDF generation are pretty limited. Agree on it not being worth the cost right now

src/routes/_authenticated/forms/social-security.tsx
kydecker: Should make it easy to have form-specific instructions for the cover pages!

PR #640, opened 2025-06-19
feat: Improve placeholder quest management

### LINKED ISSUE #355: Allow manually removing placeholder quests
Currently there are 5 placeholder quests shown to new users:

Image

- [x] Figure out the data model for placeholder quests
- [x] What happens when a user dismisses one but then wants it back
- [x] What does the dismiss button look like, is there confirmation?

Some users may have already completed particular steps, or they may not want to change a particular document. Allow dismissing these placeholders from the view.

kydecker:
## What changed?
Fixes #355. Fixes #640.

- Adds the ability to dismiss quest placeholders (for example, if you don't want to do a birth certificate change).
- Changes the flow for adding quests from a placeholder; rather than displaying an interstitial which saves a "residence" or "birthplace" to the `users` table and then uses that information to derive a proper quest slug for linking to a quest (which is flaky and error-prone), the new approach will allow the user to click on a placeholder and then display a filtered view of search results for quests within that category
- Adds the ability to search by jurisdiction name in addition to quest title
- Add an empty state when all placeholders have been dismissed and no user quests exist
- Adds ability to filter the quests search by category
- Removes "residence" and "birthplace" from users table since they are no longer needed.
- Removes user settings and components related to residence and birthplace.
- Removes `JurisdictionInterstitial` component since it's no longer needed.

https://github.com/user-attachments/assets/bd4af945-128e-405c-85d3-6a1ad4291a2a

## Why?
Previously it was impossible to dismiss placeholder suggestions, and they were hardcoded into the frontend with some logic that made assumptions about what the slug would look like for each quest. This change improves flexibility of display for placeholders, allows dismissing and restoring placeholders, and is generally easier to reason about.

A side-effect of this change is that we no longer need to store residence/birthplace on the user table. The placeholders, when clicked, change to the "all quests" tab of the sidebar and filter the view to whatever category was clicked.

CleanShot 2025-06-19 at 22 33 55@2x

In the filtered view, when viewing the categories for court order, state id, or birth certificate, the state name will display in place of the quest title, to prevent repetition of "Court Order" "Court Order" "Court Order", etc.

Also, if a court order, state ID, or birth certificate is added to the user's quests, the remaining court orders, state ids, and birth certificates will be hidden from the search pane. This is implemented via a new `quests.getRelevantActive` API endpoint, with the assumption that after a single quest is chosen for any of these categories, the remaining categories are no longer relevant.

## How was this change made?
- Created new `userQuestPlaceholders` table
- Created new model and API to write to the table
- Updated `QuestsSidebar` component with logic to fetch and display the placeholders
- Removed existing hard-coded frontend logic
- Removed `JurisdictionInterstitial` component and tests
- Removed residence and birthplace user settings components and tests

## How was this tested?
Added new tests for updated models, API, and components.


kydecker:
@mbbischoff This ended up bigger than expected. Tldr; moved a bunch of placeholder quest logic to the backend, updated the user flows for clicking from placeholder → adding a quest.

kydecker:
@mbbischoff I need to merge some of this logic to unblock other tasks, but feel free to leave comments after merge and I'm happy to address edits

PR #666, opened 2025-06-24
chore: Improve chunking and dynamically load certain imports

kydecker:
## What changed?

Split the main bundle into multiple smaller chunks that can be loaded asynchronously.

| Before | After |
|--------|--------|
| CleanShot 2025-06-24 at 14 23 42@2x | CleanShot 2025-06-24 at 14 22 01@2x |
| CleanShot 2025-06-24 at 14 23 07@2x | CleanShot 2025-06-24 at 14 21 21@2x | 

## Why?
Running `npx vite-bundle-visualizer` surfaced a warning about chunks larger than 500kb, and indeed, most of the application was contained in two large `index` chunks, one 1,216.42 kB and one 2,776.07 kB.

This unnecessarily loaded libraries like `typed-usa-states` (used for the State and County select component) and `language-name-map` (language select) along with larger utilities like `pdf-lib` even if they were not used.

## How was this change made?
1. Used dynamic imports for certain components to prevent their library being loaded until the component is invoked (`typed-usa-states`, `pdf-lib`)
2. Specified manual chunks to use when building with Vite, causing the chunks to only load if necessary instead of all being bundled and downloaded at once

## How was this tested?
All existing tests pass, manually verified that PDF downloads work.

## Anything else?
Yay performance 🚀

### DIFF
--- a/.changeset/giant-animals-thank.md
+++ b/.changeset/giant-animals-thank.md
@@ -1,0 +1,5 @@
+---
+"namesake": patch
+---
+
+Reduce initial bundle size and dynamically load utilities more intelligently to improve app speed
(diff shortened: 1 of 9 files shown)


kydecker:
Cool 😎

CleanShot 2025-06-24 at 14 55 56@2x

Issue #787, opened 2025-09-23
"This modal is already wrapped in a ModalOverlay, props [isOpen, onOpenChange] should be placed on the ModalOverlay instead."


evadecker:
**Describe the bug**
There are repeated warnings in the console when running `pnpm test`:

> This modal is already wrapped in a ModalOverlay, props [isOpen, onOpenChange] should be placed on the ModalOverlay instead.

**To Reproduce**
1. Pull changes
2. Run `pnpm test`
3. View warnings

**Expected behavior**
No warnings in the console.

**Additional context**
Relocate props from `Modal` to `ModalOverlay` as mentioned in the message.

evadecker:
Closed in #788

PR #788, opened 2025-09-23
fix: Eliminate console warnings about `Modal` and `ModalOverlay` usage

jeetsingh008:
What changed?
This PR refactors the Modal component to resolve an error from react-aria-components which fixes the issue #787. The state-management props (isOpen, onOpenChange, etc.) are now correctly passed only to the ModalOverlay component, instead of being passed to both the overlay and the inner AriaModal.

This fixes the console error: "This modal is already wrapped in a ModalOverlay, props [isOpen, onOpenChange] should be placed on the ModalOverlay instead."

Why?
The previous implementation was passing all incoming props to both ModalOverlay and AriaModal. The react-aria-components library is designed for ModalOverlay to be the sole manager of the modal's open/close state. Passing these state props to the inner AriaModal created a conflict and violated the library's intended pattern. This change brings the component into alignment with the library's official usage guidelines.

How was this change made?
The fix was implemented by making a small but crucial adjustment to the Modal component:

The {...props} spread was removed from the <AriaModal> component.

The children prop was destructured from the Modal's arguments and passed explicitly to <AriaModal>.

This ensures that state-related props are exclusively handled by <ModalOverlay>, while <AriaModal> only receives the props it needs for rendering its content and styles.

How was this tested?
I confirmed the fix by running the application and performing the following checks:

Verified that the modal opens and closes correctly.

Checked the browser's developer console and ran pnpm test to ensure the original error no longer appears.

Confirmed that all modal content and styling render as expected.

### DIFF
--- a/src/components/common/Modal/Modal.tsx
+++ b/src/components/common/Modal/Modal.tsx
@@ -37,7 +37,6 @@ export function Modal({ className, ...props }: ModalOverlayProps) {
     <ModalOverlay {...props} className={overlayStyles}>
       <Dialog>
         <AriaModal
-          {...props}
           className={composeRenderProps(className, (className, renderProps) =>
             modalStyles({ ...renderProps, className }),
           )}


kydecker (approved):
Love an easy fix! Appreciate your writeup in the PR description, too.

Thanks for your contribution!

jeetsingh008:
Thank you for reviewing and merging my PR, please do tell me if there is anything else i can help with.

PR #806, opened 2025-10-31
Draft: Add NY UCSNC-1 name/sex change petition form PDF

mbbischoff:
Progress on #805.

Added the official New York UCSNC-1 petition form for individual adult name and/or sex designation change to the forms directory.

## What changed?
Before I write the TypeScript, I just wanted to have @evadecker take a look at the first form, as I annotated it, and made sure I’m correctly following the guidelines set out in the forms README.


kydecker:
Downloaded and gave a click through the form, the labels look good! Nice.

I will say that for the forms I wired up in the past it's been helpful to rely on test-driven development. So writing out the test cases you want to verify (fill out all these fields by default, fill out these fields if this conditional is checked, etc.) and then writing the TypeScript logic which gets them to pass.

